PR #5, opened 2017-01-23
Added src directory resolving on Webpack config

felipegarcia92:
Okay, so I added this and personally, I think the code looks much better using it.

However, we have a couple of situations to solve before we merge this into master. 

1) The first one has to do with the linter triggering the `import/no-unresolved` rule exception on these imports. Here's a related link with a discussion: https://github.com/AtomLinter/linter-eslint/issues/610. I tried some of the suggested methods there, but none of them work.
Also, I want to add, this doesn't have anything to do with Atom since the exceptions are also triggered when running `npm run lint` via CLI.
One possibility would be adding `'import/no-unresolved' :0` to the project's `.eslintrc`, but you may not want to do that. I don't think that's a big deal, but something to discuss.

2) Do we have a AWS bucket for this base to test it? It should work everywhere if we resolve modules using Webpack, as we're doing, but if we want an extra test on AWS, we'll need to set that up, and probably a generic API to connect to.

What's your take on this?

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -1,6 +1,6 @@
 import * as types from './actionTypes';
-import sessionApi from '../api/sessionApi';
-import * as session from '../services/sessionService';
+import sessionApi from 'api/sessionApi';
+import * as session from 'services/sessionService';
 import { SubmissionError } from 'redux-form';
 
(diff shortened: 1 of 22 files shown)


bernabe9:
#85 Resolve the issue

PR #86, opened 2018-10-04
Add the library classnames to concat css classes

### LINKED ISSUE #87: Integrate with classnames library
We should add https://github.com/JedWatson/classnames for conditional classnames and classnames in general.

bernabe9:
Resolves #87

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
   "author": "Bernabe Gonzalez",
   "license": "MIT",
   "dependencies": {
+    "classnames": "2.2.6",
     "foundation-sites": "6.4.4-rc1",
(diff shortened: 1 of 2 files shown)


bernabe9:
@bernabe9 @aguscha333 @missx @juanlacruz what do you guys think about adding this library? I usually see messy code due to the classnames concatenation.

felipegarcia92:
I like it, but I'm starting to think we need some basic styling in the base to make use of it and kind of show to new devs how to use it.

bernabe9:
@felipegarcia92 I agree. I'd prefer to have a document about how to do styling with examples & FAQ.

Issue #97, opened 2018-12-17
Go to login when api returns 401 unauthorized


aguscha333:
### Problem
Users can continue to navigate the site if they logged out on another tab. This is the use case that i found where this happens.

The problem in the code is that in that case session can't be loaded since it does not exist anymore so it ends up in the catch and does nothing:
```
    sessionService.loadSession()
      .then(() => {
        if (response.status === 401) {
          sessionService.deleteSession();
          window.location = routes.login;
        }
      }).catch(() => {});
```

### Solution
This is the solution I came up with in my project, if we agree that this is a good solution I can create a PR to add it to the base.
```
    sessionService.loadSession()
      .then(() => {
        if (response.status === 401) {
          sessionService.deleteSession();
          window.location = routes.login;
        }
      }).catch(() => {
        window.location = routes.login;
      });
```

bernabe9:
I only see one downside to your workaround. What happens if the user isn't navigating to an authorized page? We will redirect it to the login page anyways :(

I'd prefer something like this:

```javascript
sessionService.loadSession()
  .then(() => {
    if (response.status === 401) {
      sessionService.deleteSession();
      window.location = routes.login;
    }
  }).catch(() => {
    if (response.status === 401) {
      window.location = routes.login;
    }
  });
```

If the user isn't receiving a 401 means that the user is allowed to navigate to the page.

fsarachu:
I agree there is a problem, but that solution would cause a login page reload on a failed login too, not good.
I think the main problem is that the redirect is being handled imperatively here.
How about handling it declaratively with react-router, inside the ProtectedRoute component?

bernabe9:
Issue fixed here: https://github.com/rootstrap/react-redux-base/pull/99

PR #98, opened 2018-12-17
401 response redirect handling

fsarachu:
Solves #97 
Always delete session on 401 response, this will cause session state to be refreshed and the redirect will be handled declaratively by PrivateRoute component

### DIFF
--- a/src/api/apiService.js
+++ b/src/api/apiService.js
@@ -28,13 +28,9 @@ const handleErrors = response =>
       return;
     }
 
-    sessionService.loadSession()
-      .then(() => {
-        if (response.status === 401) {
-          sessionService.deleteSession();
-          window.location = routes.login;
-        }
-      }).catch(() => {});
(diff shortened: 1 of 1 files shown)


fsarachu:
... Investigating tests failure 🤔

src/api/apiService.js
```diff
@@ -28,13 +28,9 @@ const handleErrors = response =>
...
     }
 
-    sessionService.loadSession()
-      .then(() => {
-        if (response.status === 401) {
-          sessionService.deleteSession();
-          window.location = routes.login;
-        }
-      }).catch(() => {});
+    if (response.status === 401) {
```
juanlacruz: @bernabe9 _loadSession()_ function only returns the current session if it exists or does it save the user too?

fsarachu:
Closing this so there's only one PR to solve this issue (#99)

PR #99, opened 2018-12-17
fix: logout on 401

aguscha333:
# fix: logout on 401
## Description
This PR addresses issue #97.

Problem was when session was not present and the promise entered the catch without returning to login page.

I added `refreshFromLocalStorage` before each request to make sure that the session is up to date with the storage.


src/api/apiService.js
```diff
@@ -28,13 +28,13 @@ const handleErrors = response =>
...
+    if (response.status === 401) {
     sessionService.loadSession()
       .then(() => {
-        if (response.status === 401) {
-          sessionService.deleteSession();
-          window.location = routes.login;
-        }
+        sessionService.deleteSession();
       }).catch(() => {});
+      window.location = routes.login;
```
fsarachu: If the user is already on the login page, and tries to log in with the wrong credentials, wouldn't this will cause a full page reload?

aguscha333: great catch, fixed

fsarachu:
I think this is enough to solve the current issue, but doesn't aim the root cause. 
The real problem here is that the app uses Redux as the source of truth for the session state (e.g the `PrivateRoute` component), while the library `redux-react-session` uses the storage as the source of truth, so that is the real issue that should be addressed.
If `redux-react-session` used Redux as source of truth, no `window.location` redirect would be needed, and code would be much cleaner, with the redirect being handled declaratively by `PrivateRoute`. 
Would like to to wait for @bernabe9 thoughts before merging this.

aguscha333:
@bernabe9 any thoughts on what @fsarachu said? I have some time to continue working on this.

bernabe9:
@fsarachu I agree that we can solve it declaratively.

> If `redux-react-session` used Redux as source of truth, no `window.location` redirect would be needed, and code would be much cleaner, with the redirect being handled declaratively by `PrivateRoute`

`redux-react-session` syncs the storage with Redux, so if it doesn't have bugs it should be the same source of truth. So, when we delete the session it should render again the `App` component: https://github.com/rootstrap/react-redux-base/blob/master/src/components/App.js changing the routes. @aguscha333 can you confirm if we need the `window.location`? I think that once we delete the session the route isn't valid anymore.

Anyways, I think that a `windows.location` redirection isn't a bad solution because it's not only for routes, it's for any API call.

aguscha333:
@bernabe9 One of the use cases of this issue is if you have 2 tabs open and log out in one of them, then you should be logged out in the other when you go back to it.

Do you know if `redux-react-session` syncs redux with the storage if the storage is changed?
I think that is the core issue here. To test I added an about page and marked it as private. At first I tested with no requests done to the backend and although it's a private route, since there is no request to the backend there is no 401 error that kicks the user out. 

After that I added a `getProfile` request to the backend to test a page that does make a request and gets `401 error` from the backend. With that I tested the removal of `window.location` but it didn't work. Which also makes me wonder why it didn't get kicked out by the `PrivateRoute`. Isn't the `session` reducer being updated with `sessionService.deleteSession()`?

In any case if we are able to sync redux with the storage I think that the manual redirection that

bernabe9:
> @bernabe9 One of the use cases of this issue is if you have 2 tabs open and log out in one of them, then you should be logged out in the other when you go back to it.

This use case is weird. Can you give me a real app example that is doing this? I tested on Facebook and it keeps the other tab logged in until you do a new request.

@aguscha333 did you check Redux store logs to confirm if `sessionService.deleteSession()` was updating the store? If is updating the store, check if is trying to re-render the routes.

PR #100, opened 2018-12-17
Update node version to current LTS

aguscha333:
# Update node version to current LTS

## Description
This PR is to address the error that sometimes appears when running `yarn` or `npm install` for the first time when your version of node is not the one in `package.json`.

### Error
```
yarn install v1.10.1
$ node tools/nodeVersionCheck.js
[1/5] :mag:  Validating package.json...
error react_redux_base@0.1.0: The engine "node" is incompatible with this module. Expected version "9.8". Got "10.13.0"
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

By updating the version to the current LTS is less likely that the developers using the base will encounter the error. I also updated the readme file with a small explanation of what version to use or what to do if you want to use a different one.


fsarachu:
Node version should be also bumped in `.nvmrc` file

PR #102, opened 2019-01-09
add redux selectors

### LINKED ISSUE #96: Add redux selectors
It would be great to have memoized selectors to compute derived data from the redux store https://redux.js.org/recipes/computingderiveddata. Not only for computing derived data, but also to stop doing toJS calls on the mapState function, that are triggering re-renders because of the nature of the toJS method, and the shallow compare that React-Redux does

pedrozs:
# add redux selectors
## Description
This PR resolves #96.
It also adds the username to the homepage message, to give an example of a selector.

### DIFF
--- a/src/containers/HomePage.js
+++ b/src/containers/HomePage.js
@@ -1,13 +1,24 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
+import { object } from 'prop-types';
+import { connect } from 'react-redux';
 
 import LogoutButton from 'components/user/LogoutButton';
+import { getUser } from 'selectors/sessionSelectors';
 
-const HomePage = () => (
+const HomePage = ({ user }) => (
   <div>
-    <p><FormattedMessage id="home.welcome" /></p>
+    <p><FormattedMessage id="home.welcome" values={user} /></p>
     <LogoutButton />
   </div>
 );
 
-export default HomePage;
+HomePage.propTypes = {
+  user: object
+};
+
+const mapState = state => ({
+  user: getUser(state)
+});
+
+export default connect(mapState)(HomePage);
--- a/src/locales/es.js
+++ b/src/locales/es.js
@@ -19,7 +19,7 @@ export default {
 
   'logout.button': 'LOGOUT',
 
-  'home.welcome': 'Welcome to React Redux Base',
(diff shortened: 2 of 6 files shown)


src/containers/HomePage.js
```diff
@@ -1,13 +1,18 @@
...
+    <p><FormattedMessage id="home.welcome" />{` ${username}`}</p>
     <LogoutButton />
   </div>
 );
 
-export default HomePage;
+const mapState = state => ({
+  username: getUser(state).username
+});
+export default connect(mapState)(HomePage);
```
bernabe9: add an empty line above

src/containers/HomePage.js
```diff
@@ -1,13 +1,18 @@
...
+import { connect } from 'react-redux';
 
+import { getUser } from 'selectors/sessionSelectors';
 import LogoutButton from 'components/user/LogoutButton';
 
-const HomePage = () => (
+const HomePage = username => (
   <div>
-    <p><FormattedMessage id="home.welcome" /></p>
+    <p><FormattedMessage id="home.welcome" />{` ${username}`}</p>
```
bernabe9: Why are you adding a space before the username? 🤔

pedrozs: To add a space between the welcome message and the username. I've added it as a value now anyway.

src/locales/es.js
```diff
@@ -19,7 +19,7 @@ export default {
 
   'logout.button': 'LOGOUT',
 
-  'home.welcome': 'Welcome to React Redux Base',
+  'home.welcome': 'Welcome to React Redux Base {username}',
```
missx: Shouldn't this be in Spanish?

src/containers/HomePage.js
```diff
@@ -1,13 +1,19 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
+import { connect } from 'react-redux';
 
+import { getUser } from 'selectors/sessionSelectors';
 import LogoutButton from 'components/user/LogoutButton';
 
-const HomePage = () => (
+const HomePage = username => (
```
fsarachu: Aren't you missing the destructuring `({username})`? 🤔

src/locales/es.js
pedrozs: It is now.

PR #103, opened 2019-01-09
fix s3 upload

### LINKED ISSUE #94: Fix sync when uploading to s3
screen shot 2018-11-12 at 9 43 45 pm
Hi!, I think that maybe when uploading to s3, it should wait until all promises are fulfilled instead of waiting just the last one.

peter7z:
# Fix s3 upload
This PR fixes #94 
S3 client methods do not return a promise, so I had to construct them.
Suggestions are very welcome.

### DIFF
--- a/tools/s3.js
+++ b/tools/s3.js
@@ -14,76 +14,75 @@ class s3 {
     this.bucket = process.env.AWS_BUCKET;
   }
 
-  clearBucket() {
+  // Private
+  readdir(folderPath) {
     return new Promise((resolve, reject) => {
-      this.client.listObjects({ Bucket: this.bucket }, (err, data) => {
-        if (err) {
-          reject(`error listing bucket objects ${err}`);
-          return;
+      fs.readdir(folderPath, (err, files) => {
+        if (!files || files.length === 0) {
+          reject(new Error('provided folder is empty or does not exist.'));
         }
-        
-        const items = data.Contents;
-
-        if (items.length === 0) {
-          resolve();
-        }
-
-        items.forEach((item, index) => {
-          const deleteParams = { Bucket: this.bucket, Key: item.Key };
-          this.client.deleteObject(deleteParams, (err, data) => {
-            if (err) {
-              reject(`delete err ${deleteParams.Key}`);
-            }
-            if (index === items.length - 1) {
-              resolve();
-            }
-          })
-        })
-      })
-    })
+        if (err) reject(err);
+        else resolve(files);
+      });
+    });
   }
 
-  syncDir(distPath) {
-    return new Promise((resolve, reject) => {
-      const folderPath = path.join(__dirname, distPath);
+  // Private
+  uploadFile = fileName => new Promise((resolve, reject) => {
+    const filePath = path.join(this.folderPath, fileName);
+    fs.readFile(filePath, (error, fileContent) => {
+      if (error) reject(error);
+      const s3Params = {
+        Bucket: this.bucket,
+        Key: fileName,
+        Body: fileContent,
+        ACL: 'public-read',
+        ContentType: mime.getType(fileName)
+      };
 
-      fs.readdir(folderPath, (err, files) => {
-        if(!files || files.length === 0) {
-          reject(`provided folder '${distPath}' is empty or does not exist.`);
-          return;
+      if (/\.js$|\.css$/.test(fileName)) {
+        s3Params.ContentEncoding = 'gzip';
+      }
+
+      this.client.putObject(s3Params, (err, data) => {
+        if (err) {
+          console.log(`Error uploading ${fileName} ${err.code}`);
+          return reject();
         }
 
-        files.forEach((fileName, index) => {
-          const filePath = path.join(folderPath, fileName);
-
-          fs.readFile(filePath, (error, fileContent) => {
-            if (error) { reject(error); }
-
-            const s3Params = {
-              Bucket: this.bucket,
-              Key: fileName,
-              Body: fileContent,
-              ACL: 'public-read',
-              ContentType: mime.getType(fileName)
-            };
-            if (/\.js$|\.css$/.test(fileName)) {
-              s3Params.ContentEncoding = 'gzip';
-            }
-
-            // upload file to S3
-            this.client.putObject(s3Params, (err) => {
-              if (err) {
-                reject(err)
-              } else {
-                if (index === files.length - 1) {
-                  resolve();
-                }
-              }
-            });
-          });
-        });
+        return resolve(data);
       });
-    })
+    });
+  })
+
+  // Private
+  deleteItem = (item) => {
+    const deleteParams = { Bucket: this.bucket, Key: item.Key };
+    return new Promise((resolve, reject) =>
+      this.client.deleteObject(deleteParams, (error) => {
+        if (error) return reject(error);
+        return resolve();
+      }));
+  };
+
+  async clearBucket() {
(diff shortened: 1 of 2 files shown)


tools/s3.js
```diff
@@ -21,24 +21,25 @@ class s3 {
...
-            }
-            if (index === items.length - 1) {
-              resolve();
-            }
-          })
+          deleteItemsPromises.push(this.client.deleteObject(deleteParams));
         })
+
+        Promise.all(deleteItemsPromises)
+          .then(() => {
```
bernabe9:
```suggestion
          .then(resolve)
          .catch((err) => reject(`delete err ${deleteParams.Key}`)
```

tools/s3.js
```diff
@@ -70,18 +73,16 @@ class s3 {
...
-                  resolve();
-                }
-              }
-            });
+            uploadItemsPromises.push(this.client.putObject(s3Params));
           });
         });
+
+        Promise.all(uploadItemsPromises)
+          .then(() => {
```
bernabe9:
```suggestion
          .then(resolve).catch(reject)
```

tools/s3.js
```diff
@@ -21,24 +21,22 @@ class s3 {
...
-        
-        const items = data.Contents;
 
+        const items = data.Contents;
         if (items.length === 0) {
           resolve();
         }
 
+        const deleteItemsPromises = [];
         items.forEach((item, index) => {
```
felipegarcia92: Can you make this an Array.map?

tools/s3.js
```diff
@@ -53,35 +51,38 @@ class s3 {
           return;
         }
 
+        const uploadItemsPromises = [];
+
         files.forEach((fileName, index) => {
```
felipegarcia92: Same here, normally `forEach + push = map`

tools/s3.js
```diff
@@ -53,35 +51,38 @@ class s3 {
...
+                ContentType: mime.getType(fileName)
+              };
+              if (/\.js$|\.css$/.test(fileName)) {
+                s3Params.ContentEncoding = 'gzip';
               }
+
+              this.client.putObject(s3Params,(err, data) => {
+                if (err) console.log(`Error uploading ${fileName}`, err.code);
+                else console.log(data);
+                resolve();
```
missx: You are always resolving even thought putObject can throw an error?

tools/s3.js
```diff
@@ -21,24 +21,22 @@ class s3 {
...
         items.forEach((item, index) => {
           const deleteParams = { Bucket: this.bucket, Key: item.Key };
-          this.client.deleteObject(deleteParams, (err, data) => {
-            if (err) {
-              reject(`delete err ${deleteParams.Key}`);
-            }
-            if (index === items.length - 1) {
-              resolve();
-            }
+          const promise = new Promise((resolve,reject) => {
```
missx: Could you add a space between ```resolve,``` and ```reject```?

tools/s3.js
```diff
@@ -21,24 +21,22 @@ class s3 {
...
+            this.client.deleteObject(deleteParams, resolve)
```
missx: And can you put a semicolon here?

tools/s3.js
bernabe9: remove `reject` param, you're not using it.

```suggestion
          const promise = new Promise((resolve) => {
```

tools/s3.js
```diff
@@ -21,24 +21,22 @@ class s3 {
           reject(`error listing bucket objects ${err}`);
           return;
         }
-        
-        const items = data.Contents;
 
+        const items = data.Contents;
         if (items.length === 0) {
           resolve();
```
fsarachu: You should do `return resolve();`, function execution will continue if not, probably not what you want

tools/s3.js
bernabe9: You should reject the promise when `deleteObject` function fails.

```suggestion
            this.client.deleteObject(deleteParams,(error) => {
              if (error) {
                reject(`delete err ${deleteParams.Key}`);
              } else {
                resolve();
              }
            })
```

tools/s3.js
```diff
@@ -21,24 +21,22 @@ class s3 {
...
-            if (index === items.length - 1) {
-              resolve();
-            }
+          const promise = new Promise((resolve,reject) => {
+            this.client.deleteObject(deleteParams, resolve)
           })
+          deleteItemsPromises.push(promise);
         })
+        Promise.all(deleteItemsPromises)
+          .then(resolve).catch((err) => reject(`delete err ${deleteParams.Key}`))
```
bernabe9: You don't have access to `deleteParams`. Just return the error

tools/s3.js
```diff
@@ -53,35 +51,38 @@ class s3 {
...
-              Body: fileContent,
-              ACL: 'public-read',
-              ContentType: mime.getType(fileName)
-            };
-            if (/\.js$|\.css$/.test(fileName)) {
-              s3Params.ContentEncoding = 'gzip';
-            }
+          const promise = new Promise((resolve, reject) => {
+            fs.readFile(filePath, (error, fileContent) => {
+              if (error) { reject(error); }
```
fsarachu: Same return thing. In this case code will continue and try to upload objects even if there was an error reading the file.

pedrozs:
What do guys think? It seems to me that it's still kinda hard to read.

fsarachu:
> I still think it's still kinda hard to read... What do you guys think?

Yeah, also, I don't think promises is the best option for this kind of job. For example, when an upload fails, the other promises are left there hanging and we have no control over them because of the `Promise.all()` fail fast behavior.
I think this script deserves a full rewrite.

pedrozs:
Do you have any suggestions @fsarachu?
> when an upload fails, the other promises are left there hanging

I can't think of another implementation besides a synchronous one.

fsarachu:
> Do you have any suggestions @fsarachu?
> 
> > when an upload fails, the other promises are left there hanging
> 
> I think possible solutions could be:
> 
> * A synchronous implementation
> * Something like bluebird that lets you cancel promises.

Keeping promises, we could do [some trick to avoid the fail-fast behavior](https://nmaggioni.xyz/2016/10/13/Avoiding-Promise-all-fail-fast-behavior/).
The other option is using plain callbacks and use some state to keep track of the uploaded files (state for async stuff 👎, very prone to bugs), or we can go full crazy and try out [Observables](https://rxjs-dev.firebaseapp.com/guide/overview) as this could be accomplished very easy with them (probably not worth the overhead).

Taking everything in mind I would go with the first one, doing that promise trick, but I would like to hear other's thoughts.

PS: Of course the promises option can be sweetened with some async/await sugar, but that's another story.

agustito37:
Alo! You can transform s3 methods to Promises I think, IDK if it's worth but it could make the code more readeable if you add async/await syntax.
The fast-fail avoiding link as @fsarachu wrote seems to be a good solution, but I have a concern about the consistency in this script. Shouldn't this has like a retry mechanism or something if a Promise of the list fails?

fedeagripa:
@peter7z can we either merge this or close it?

peter7z:
@fedeagripa I've made some requested changes and then nobody checked again. This works, and works better than what theres in the code right now. It's still far from an optimal solution though.
It can be either closed or merged. what do you think?

fedeagripa:
@agustito37 @bernabe9 as you have still some request changes, can we merge this as it is ? or there is still something else to do before merging?

PR #107, opened 2019-01-14
fix circleCi not restoring cache

### LINKED ISSUE #91: Circle CI cache not configured
Circle CI `config.yml` is caching `yarn` `cache` but is not restoring it at the beginning of the script.

More info: https://circleci.com/docs/2.0/yarn/

Probably addressed in cypress integration.

pedrozs:
This PR resolves #91 
Cache was being saved, but never restored.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,10 @@ jobs:
 
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


pedrozs:
@agustito37 Thank you for the help!

PR #111, opened 2019-03-25
Update redux devtools

nsantos16:
### Description
window.devToolsExtension is deprecated in favor of window.\_\_REDUX_DEVTOOLS_EXTENSION__ , and will be removed in next version of Redux DevTools: https://git.io/fpEJZ


nsantos16:
> 🎉 Thanks for making the update, is this change something that we should be implementing in existing projects?

Not critical at all, is just a fix since it will be a deprecated method.

PR #112, opened 2019-04-27
Update sass to version 4.12

peter7z:
Last version had a problem in install script so I had to update.
4.12 is the latest stable version and it supports node 12.


mariosouto:
@peter7z Shouldn't we update the yarn.lock also?

peter7z:
> @peter7z Shouldn't we update the yarn.lock also?

Yeah sorry, I use npm

mariosouto:
👌

PR #113, opened 2019-05-23
Fix tests failing without API_URL environemnt variable

peter7z:
Tests were failing because nock couldn't find `process.env.API_URL`.
Ideas for a better fix are welcome!

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+API_URL=https://mock.com/api/v1
--- a/tools/testSetup.js
+++ b/tools/testSetup.js
@@ -1,6 +1,6 @@
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
-require('dotenv').config({ path: '.env.dev'});
+require('dotenv').config({ path: '.env.test'});
 
 Enzyme.configure({ adapter: new Adapter() });


peter7z:
@felipegarcia92 If there is a .env.dev that line will load it, the line added in the PR is there just in case there is no .env or that it doesn't specify an API_URL variable.

fsarachu:
How about having a `.env.test` file with mocked values instead?
So it would just be changing that line to `require('dotenv').config({ path: '.env.test'});`
Makes sense to me having a different .env file for tests

kevquincke:
Agree with @fsarachu currently we were ignoring `.env.dev` that's probably the reason that it was failing. The best thing to do is:
1. Keep `reequire('dotenv').config({ path: '.env.test'});`
2. Add a new .env.test file and push it

We also need to make sure this project is compiling on CircleCI without depending on any env var on CircleCI

mariosouto:
Agree also with @fsarachu 

But we could load a different `.env` file based on the environment without having to do this on the testSetup.

Something like a generic

```js
require('dotenv').config({ path: process.env === 'dev' ? '.env.dev' : 'env.test' })
```

how do you guys feel about it?

fsarachu:
> Agree also with @fsarachu
> 
> But we could load a different `.env` file based on the environment without having to do this on the testSetup.
> 
> Something like a generic
> 
> ```js
> require('dotenv').config({ path: process.env === 'dev' ? '.env.dev' : 'env.test' })
> ```
> 
> how do you guys feel about it?

We could take this further and just do this:

```js
require('dotenv').config({ path: `.env.${process.env || 'dev'}` })
```

peter7z:
@fsarachu @mariosouto @kevquincke
We decided to just load `.env.test` since when `testSetup.js` runs the environment will always be test. The `.env.test` file was also pushed.

PR #115, opened 2019-05-28
Cypress

aguscha333:
This PR integrates [Cypress](https://www.cypress.io/) into the base. For those who don't know what Cypress is, it's testing tool that allows you to easily implement not only unit tests but also integration and e2e.
 
The intention of this PR is to have a place where other developers can leave feedback on the implementation and documentation as well.

I still have some sections to complete in the readme but the most important ones have already a draft written.

Any questions that you have please let me know.

Here is a link to the [README](https://github.com/rootstrap/react-redux-base/tree/cypress/cypress) I added.

-----
### Update
Readme updated to be more complete and easy to understand.

To Do:
add support for e2e on the frontend side.


cypress.json
```diff
@@ -0,0 +1,12 @@
+{
+  "baseUrl": "http://localhost:8000",
+  "indexUrl": "/",
+  "env": {
+    "API_URL": "https://test.com/api/v1"
```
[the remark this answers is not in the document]
aguscha333: I was able to configure cypress to use the env defined in process.env so it does not need to be repeated.

src/utils/history.js
```diff
@@ -8,6 +8,9 @@ class History {
     if (!instance) {
       instance = process.env.BROWSER ? createBrowserHistory() : createMemoryHistory();
     }
+    // if (window.Cypress) {
```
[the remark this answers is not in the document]
aguscha333: Nice catch! actually it was part of an attempt to solve something that didn't work so I'll remove it.

cypress/integration/containers/LoginPage.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    it('submit successful, should be redirected to the homepage', () => {
+      cy.stubRequest(loginStub());
+
+      cy.get('input[name="email"]').type(email);
+      cy.get('input[name="password"]').type(password);
+      cy.get('form')
+        .submit()
+        .wait('@loginStub');
+
+      cy.visit('/login');
```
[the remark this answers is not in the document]
aguscha333: Hmm I think that what it's actually being tested here is that after you logged in if you try going back to the login page you are always redirected to the home page.

Perhaps I should also assert that before that visit the url is also the homepage so we also test that after login user is redirected to the homepage.

At the same time it's kind of the same bit of code that redirects the user so we would be testing it twice.

```
  if (authenticated) {
    return <Redirect to={routes.index} />;
  }
```

WDYT?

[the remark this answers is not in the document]
felipegarcia92: I think I missed the test unit description, which already explains what you just explained in the comment. This was about the redirection, so it looks good. I don't think you need to extend this test or anything.

cypress/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+As you can see the function is receiving a parameter called `customUser`. This can come in handy if for a specific case in one call you need to customize the response or any other attribute.
+
+#### Basic attributes
+
+
+ Parameter | Description | Values / Type
+ --------- | ----------- | -------------
+ name | this will be the stub identifier | **string**
+ method | Show file differences that haven't been staged | **HTTP METHODS** defined in constants
```
santiestra: Is the description for `method` ok?

aguscha333:
@santiestra regarding leaving Jest integration, I didn't remove jest completely. What I removed were the specs that we had since now those flows are tested in cypress. I also removed the scripts in package.json and replaced them with cypress ones.
I can add a script that is for example : `test:jest` that runs the old command so cypress is still the default.
I think that having the same test command run both is something we should take into consideration if we see that most projects end up using both. If not I think it's better to keep Jest as a separate command.

cypress/README.md
aguscha333: 🙈 no hehe, copy & paste did me wrong, I'll fix it.

cypress/README.md
```diff
@@ -0,0 +1,300 @@
...
+    - [Defining fields](#defining-fields)
+      * [Basic attributes](#basic-attributes-1)
+      * [Validation structure](#validation-structure)
+      * [Validation options](#validation-options)
+
+
+----
+
+## Setup
+Cypress comes almost ready for you to use in this branch. There are a few configurations that you need to change in order to get the most out of it.
```
bernabe9:
```suggestion
Cypress comes almost ready for you to use it. There are a few configurations that you need to change in order to get the most out of it.
```

Since we are going to merge this branch, I wouldn't mention it

package.json
```diff
@@ -87,17 +92,22 @@
     "compression-webpack-plugin": "1.1.11",
     "connect-history-api-fallback": "1.5.0",
     "css-loader": "1.0.0",
+    "cypress": "3.3.0",
+    "cypress-image-snapshot": "3.0.2",
+    "cypress-react-unit-test": "2.4.1",
```
bernabe9: where are you using this package?

package.json
```diff
@@ -87,17 +92,22 @@
...
     "dotenv": "6.0.0",
     "dotenv-webpack": "1.5.7",
     "enzyme": "3.4.1",
     "enzyme-adapter-react-16": "1.5.0",
     "eslint": "4.19.1",
     "eslint-config-airbnb": "16.1.0",
+    "eslint-plugin-cypress": "2.2.1",
```
bernabe9: where are you using this package?

cypress/README.md
```diff
@@ -0,0 +1,300 @@
...
+* **Set record key:** Cypress has a handy feature called Cypress dashboard, there you can see video recordings of failed tests run in the CI of your choice.
+
+  It's very easy to get it configured, here is a [link](https://docs.cypress.io/guides/core-concepts/dashboard-service.html#Setup) to the official documentation with the steps you need to follow.
+
+* **Set `API_URL`:** Set values for `API_URL` in `.env.test_integration` and `.env.test_e2e` with the backend url you are going to use in each. If you are not going to test in one of those modes there is no need to fill it.
+
+-----
+
+## Running tests
+Running tests  is very easy, and we have already included some for you. To get a taste of how cypress works start your server with `ENV=cypress yarn start` and then run `yarn test:open`. The Test Runner will open and a list of tests to run will appear. Choose one of them and voilá, you will see screenshots of the test running in real time.
```
missx: Two spaces after Running tests?

cypress/README.md
```diff
@@ -0,0 +1,300 @@
...
+  If E2E is the way yo want to go, you won't need to do stubs to the backend. That being said, you will need to have a proper env setup in the backend for testing and use `cy.realLoginUser` instead of `cy.loginUser` to make sure that the user is authenticated from the backend's perspective.
+
+### Stubbing requests
+To stub any request you can use the custom command `cy.stubRequest` (defined in `support/commands`).
+
+You can see examples of how it's used in all of the provided specs.
+
+For each stub you want to do, you will need to define certain parameters you want to pass to it.
+
+To keep stubs organized, there is a folder call `stubs` that should be the equivalent to `api` folder in `src`.
```
missx: `called`

cypress/README.md
```diff
@@ -0,0 +1,300 @@
...
+##### Basic attributes
+ Parameter | Description | Values / Type | Required
+ --------- | ----------- | ------------- | --------
+ title | This is to show in the test case definition and to define default validation messages.  | **string** | **true**
+ name | Input name | **string** | **true**
+ inputType | Input type, used to identify element. By default the defined `inputTypes` are `INPUT` and `TEXTAREA`, you can expand that to fit your needs. | **string**, default value: `INPUT` from `inputTypes` | **false**
+ errors | Define errors to check. | **array of object**, objects inside follow **validation structure** | **false**
+ warnings | Define warnings to check. | **array of object**, objects inside follow **validation structure** | **false**
+
+Inside errors and warnings arrays you can define validations with this attributes:
```
missx: Here attributes is plural, so `these` instead of `this`

package.json
aguscha333: I am not. I added it when I was trying stuff and thought oh cool! you can write unit tests if you use this library.

I think it's definitely worth taking a look at it later but will remove it for now.
Perhaps we can write a unit test for the input component, we don't have many components in the base to showcase how it could be used.

cypress/README.md
aguscha333: 🙈

package.json
aguscha333: in eslintrc
```
  "plugins": [
    "react",
    "react-hooks",
    "cypress"
  ],
```

cypress/README.md
aguscha333: Fixed, I switched it up a bit.

PR #116, opened 2019-05-31
Refactor Api Service into common methods

mariosouto:


src/api/utils/handleErrors.js
```diff
@@ -0,0 +1,33 @@
+import { sessionService } from 'redux-react-session';
+import saveSessionHeaders from './saveSessionHeaders';
+
+export default response =>
+  new Promise((resolve, reject) => {
```
fsarachu: Lot of stuff is going on here, I think we can come up with a cleaner solution using an `async` function. Also, errors are non consistent, some returned promises are ignored, the whole function isn't always returning when a reject or resolve is called, making this really hard to read and probably easy to break.

How about something like this instead? (Didn't test this, just a base for getting thoughts on it)

```js
export default async (response) => {
  if (!response) {
    throw new Error('No response returned from fetch');
  }

  await saveSessionHeaders(response.headers);

  if (response.ok) {
    return response;
  }

  if (response.status === 401) {
    try {
      if (await sessionService.loadSession()) {
        sessionService.deleteSession();
        sessionService.deleteUser();
      }
    } catch (e) {
      console.error(e); // Or just do nothing?
    }
  }

  throw await response.json()
    .then(json => json || new Error(response.statusText))
    .catch(() => new Error('Response not JSON'));
};

File: src/api/utils/handleErrors.js [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+
+  await saveSessionHeaders(response.headers);
+
+  if (response.ok) {
+    return response;
+  }
+
+  if (response.status === 401) {
+    try {
+      await sessionService.loadSession()
```
bernabe9: `await` and `then`?

```suggestion
      await sessionService.loadSession();
      sessionService.deleteSession();
      sessionService.deleteUser();
```

src/api/utils/handleErrors.js
```diff
@@ -0,0 +1,29 @@
...
+        .then(() => {
+          sessionService.deleteSession();
+          sessionService.deleteUser();
+        });
+    } catch (e) {
+    }
+  }
+
+  throw await response.json()
```
bernabe9: I don't understand this line, are you throwing an error even if the `response.json()` is valid?

PR #118, opened 2019-06-03
Replace immutable with immer

peter7z:
Most files are just removing immutable. Important files are `createReducer` and the reducers to see how it works.

Pros taken from immer page:

   * Immutability with normal JavaScript objects and arrays. No new APIs to learn!
   * Strongly typed, no string based paths selectors etc.
   * Structural sharing out of the box
   * Object freezing out of the box
   * Deep updates are a breeze
   * Boilerplate reduction. Less noise, more concise code.
   * Small size

Cons:
 * I guess we loose the performance boost of deep object comparison of `Immutable` (we never did that anyway...)
 * You tell me!


package.json
```diff
@@ -39,7 +39,7 @@
     "foundation-sites": "6.4.4-rc1",
     "history": "4.7.2",
     "humps": "2.0.1",
-    "immutable": "3.8.2",
+    "immer": "3.1.2",
```
mariosouto: Don't forget to commit the `yarn.lock` also

server/_app.js
```diff
@@ -36,8 +36,8 @@ App.propTypes = {
 };
 
 const mapState = state => ({
-  checked: state.getIn(['session', 'checked']),
-  authenticated: state.getIn(['session', 'authenticated'])
+  checked: state.session.checked,
```
mariosouto: what do you think about using selectors for performing the computations in all the `mapState`?

peter7z: I'd love to, yes. That's for another PR though.

package.json
peter7z: My bad, I will.
Should we though? I think this is a topic for discussion. As a boilerplate project, shouldn't the versions be less restricted? As in a library?

peter7z:
@bernabe9 Done!

PR #119, opened 2019-06-07
Add a custom hook

peter7z:
I want to incentive the use of hooks. I don't know if this hook in the base will accomplish anything but here it goes. What do you think?

Here is a working example that uses this hook.
https://codesandbox.io/s/react-hooks-toggle-bq2b6?fontsize=14


bernabe9:
I agree with @felipegarcia92, I think we shouldn't have unused code in the base. We can keep the aliases though!

@peter7z great job on researching hooks! I can't wait to see hooks in real projects!

PR #120, opened 2019-06-10
Replace isomorphic fetch for axios

mariosouto:


src/api/apiService.js
```diff
@@ -1,143 +1,25 @@
...
-  });
-
-const getResponseBody = (response) => {
-  const bodyIsEmpty = response.status === 204;
-  if (bodyIsEmpty) {
-    return Promise.resolve();
-  }
-  return response.json();
-};
+import httpClient from './httpClient';
```
felipegarcia92: If this file is just about wrapping httpClient's methods, I think it's not useful. Let's use those.

mcousillas6: +1 to this, I think httpClient replaces ApiService altogether

mcousillas6: Also, by proxying through ApiService, you are missing a few option parameters that axios accepts (like query params)

PR #122, opened 2019-06-12
Involve react-redux hooks

felipegarcia92:
## Description

Based on new hooks added to `react-redux`, this PR is a first attempt at using them.

Still **work in progress**, we may want to abstract some code. But you can add comments and suggestions already, please do.

Docs here: https://react-redux.js.org/next/api/hooks

## Changes

- Replaced deprecated `react-router-redux` in favor of `connected-react-router`
- Replaced `connect` calls from `react-redux` with corresponding `useSelector` and `useDispatch` calls within function components.
- Created `useSession` custom hook to return session Redux store values.
- Upgraded `react-helmet` version due to conflicts.
- Upgraded `redux-form` version due to conflicts too.
- Upgraded `react-router-dom` version due to conflicts too.


src/reducers/index.js
```diff
@@ -1,13 +1,14 @@
 import { combineReducers } from 'redux';
 import { reducer as form } from 'redux-form';
 import { sessionReducer as session } from 'redux-react-session';
+import { connectRouter } from 'connected-react-router'
 
-import router from './routerReducer';
+// import router from './routerReducer';
```
felipegarcia92: Wondering what to do with this `routerReducer` file now and if it's needed 🤔

felipegarcia92: Looking at the console logs and Redux chrome extension, the site seems to be working OK without it and redux actions seem to be normal.

felipegarcia92:
> Great job 👍 !!
> 
> Maybe a way to reuse the `const authenticated = useSelector...` (that is repeated in some files) is to create a custom hook, that maybe can be named `useIsAuthenticated()` or something similar. It would just contain the call to the `useSelector(state => state.session.authenticated)`
> What do you think about that?

Yup, agree, good idea. This was a quick first attempt with no code optimizations.

src/reducers/index.js
felipegarcia92: @bernabe9 you may know about this one

src/containers/LoginPage.js
```diff
@@ -1,40 +1,29 @@
...
+
   if (authenticated) {
     return <Redirect to={routes.index} />;
   }
 
   return (
     <div>
       <p><FormattedMessage id="login.title" /></p>
-      <LoginForm onSubmit={login} />
+      <LoginForm onSubmit={user => dispatch(login(user))} />
```
mcousillas6: I think it's better if you create a callback here to avoid unnecessary re-renders, and then pass that as a prop, as shown here https://react-redux.js.org/next/api/hooks#examples so something like:

```javascript
  const login = useCallback(
    user => dispatch(login(user)),
    [dispatch]
  );
```
and then that as the `onSubmit` prop

felipegarcia92: Great, agree! Will do.

src/containers/HomePage.js
```diff
@@ -1,25 +1,20 @@
...
-const HomePage = ({ user }) => (
-  <div>
-    {user && user.email &&
-      <p><FormattedMessage id="home.welcome" values={user} /></p>
-    }
-    <LogoutButton />
-  </div>
-);
+const HomePage = () => {
+  const user = useSelector(state => state.session.user);
```
nsantos16: Maybe add this selector to a constant file like `userSelectors.js` and import from there

src/hooks/useSession.js
```diff
@@ -0,0 +1,15 @@
+import { useSelector } from 'react-redux';
+import { createSelector } from 'reselect';
+
+const selectSession = createSelector(
```
bernabe9: I'm not sure if we need to use `reselect` in this simple selector.
Using memoized selectors for such simple operation as accessing the props it doesn't worth it.

`reselect` will save the last `state.session` value and do a shallow comparison, also it will save the result value. In the end, this selector will be more "expensive" than the operation itself.

```suggestion
const selectSession = ({ session }) => ({
  authenticated: session.authenticated,
  checked: session.checked,
  user: session.user
})
```

server/_app.js
```diff
@@ -1,43 +1,36 @@
...
-);
+const App = () => {
+  const { authenticated, checked } = useSession();
 
-App.propTypes = {
-  authenticated: bool.isRequired,
-  checked: bool.isRequired
+  return (
+    <ThemeProvider theme={theme}>
+      <Fragment>
```
agustito37:
```suggestion
      <>
```

src/hooks/useSession.js
felipegarcia92: Thanks for the explanation. Addressed now.

src/containers/LoginPage.js
felipegarcia92: @mcousillas6 addressed, thanks!

src/hooks/useSession.js
```diff
@@ -0,0 +1,9 @@
+import { useSelector } from 'react-redux';
+
+const useSession = () => useSelector(({ session }) => ({
```
juanlacruz: nice! 👏

felipegarcia92:
Tested SSR locally, working fine. I think this is safe to merge.

PR #125, opened 2019-06-26
Add createAction helper

peter7z:
This helper makes it easier to create and use actions themselves to use as createReducer object keys.
For example:

userActions.js :
```js
export loginSuccess = createAction('loginSuccess');
```

userReducer.js :
```js
import { loginSuccess } from 'actions/userActions';

export default createReducer({
    [loginSuccess]: (state) => {  state.authenticated = true };
})
```


.nvmrc
```diff
@@ -1 +1 @@
-10.15.3
+10.16.0
```
bernabe9: This version should be the same as the one in the `package.json`

bernabe9:
why this:

```javascript
import createAction from './createAction'

const loginSuccess = createAction('loginSuccess')
```

is easier than this:

```javascript
const loginSuccess = { type: 'loginSuccess' }
```

???

peter7z:
> why this:
> 
> ```js
> import createAction from './createAction'
> 
> const loginSuccess = createAction('loginSuccess')
> ```
> 
> is easier than this:
> 
> ```js
> const loginSuccess = { type: 'loginSuccess' }
> ```
> 
> ???

Because your second code snippet should actually be:
```js
const loginSuccess = (payload) => ({ type: 'loginSuccess', payload })
```

Also this doesn't implement `toString`, that enables using this action directly in the reducer as a string.

PR #126, opened 2019-06-27
Replace Redux Form with Formik

nsantos16:
## Description

Based on  [PR](https://github.com/rootstrap/ownable-web/pull/5) made by @santiestra

## Pros

- According to Dan Abramov, [form state is inherently ephemeral and local, so tracking it in Redux (or any kind of Flux library) is unnecessary](https://github.com/reduxjs/redux/issues/1287#issuecomment-175351978)
- Redux-Form calls your entire top-level Redux reducer multiple times ON EVERY SINGLE KEYSTROKE. This is fine for small apps, but as your Redux app grows, input latency will continue to increase if you use Redux-Form.
- Redux-Form is 22.5 kB minified gzipped (Formik is 12.7 kB)

## Cons

- Issues with [backend validations](https://github.com/jaredpalmer/formik/issues/150)
- New API for create validations with [Yup
](https://github.com/jquense/yup)
## Changes

- Updated README 
- Updated tests
- Change Forms for adapted Formik
- Rewrite validations using Yup


mcousillas6:
Really nice! I agree with coni about trying out using `<Formik />` instead of the Hoc.

README.md
```diff
@@ -83,7 +83,7 @@ The server is prepared to fetch data directly from the backend before rendering
 | [PostCSS](https://github.com/postcss/postcss) | Transform styles with JS plugins. Used to autoprefix CSS |
 | [Stylelint](https://github.com/stylelint/stylelint) | Modern CSS linter that helps you enforce consistent conventions and avoid errors in your stylesheets. |
 | [Redux React Session](https://github.com/bernabe9/redux-react-session) | Keep your session sync with localStorage and Redux |
-| [ReduxForm](http://redux-form.com/6.4.3/) | Redux-form works with React Redux to enable an html form in React to use Redux to store all of its state. |
+| [Formik](https://github.com/jaredpalmer/formik) | Build forms easily without depending on Redux |
```
missx: Maybe we can also add yup to this list.

src/components/user/LoginForm.js
```diff
@@ -17,44 +17,46 @@ const messages = defineMessages({
...
+  <Formik
+    onSubmit={(values, { setSubmitting, setStatus }) =>
+      handleSubmit(values, setStatus, setSubmitting)}
+    validationSchema={loginSchema}
+    initialValues={{ email: '', password: '' }}
+  >
+    {({ isSubmitting, status }) => (
+      <Form>
+        {
+          status && <strong>{status}</strong>
```
felipegarcia92: You can have this in one line.

src/components/user/LoginForm.js
```diff
@@ -17,44 +17,46 @@ const messages = defineMessages({
...
-      />
-    </div>
-    <button type="submit">
-      <FormattedMessage id="login.form.submit" />
-    </button>
-    {isSubmitting && <Loading />}
-  </Form>
+export const LoginForm = ({ handleSubmit, intl }) => (
+  <Formik
+    onSubmit={(values, { setSubmitting, setStatus }) =>
```
felipegarcia92: I'd indent it like this:
```
onSubmit={
  (values, { setSubmitting, setStatus }) => handleSubmit(values, setStatus, setSubmitting)
}
```

or if it's too long, like this: 

```
onSubmit={
  (values, { setSubmitting, setStatus }) => 
    handleSubmit(values, setStatus, setSubmitting)
}
```

or this

```
onSubmit={(values, { setSubmitting, setStatus }) => 
    handleSubmit(values, setStatus, setSubmitting)
}
```

But for me the important part is that the line that opens the bracket and the one that closes it, are in the same indentation level.
Check for the same thing on other forms.

src/actions/sessionActions.js
```diff
@@ -1,17 +1,16 @@
...
   async () => {
     try {
       const response = await sessionApi.login({ user });
       sessionService.saveUser(response.user);
     } catch (err) {
-      throw new SubmissionError({
-        _error: err.error
-      });
+      setStatus(err.error);
+    } finally {
```
bernabe9: We need to do this in all the submission functions, right? It looks like boilerplate code.

src/components/common/Input.js
```diff
@@ -5,21 +5,29 @@ import { FormattedMessage } from 'react-intl';
...
-  meta: { touched, error }
+  form: { touched, errors, status }
 }) => (
   <div>
     {label && <label>{label}</label>}
     <div>
-      <input {...input} {...{ placeholder, type }} />
-      {touched && error &&
+      <input {...field} {...{ placeholder, type }} />
+      {touched[field.name] && status && status[field.name] &&
```
bernabe9: what's the difference between `status` and `error`?

src/components/user/LoginForm.js
```diff
@@ -10,47 +10,53 @@ import {
...
-    {submitting && <Loading />}
-  </form>
+export const LoginForm = ({ handleSubmit, intl }) => (
+  <Formik
+    onSubmit={
+      (values, { setSubmitting, setStatus }) =>
+        handleSubmit(values, setStatus, setSubmitting)
+    }
+    validationSchema={loginSchema}
+    initialValues={{ email: '', password: '' }}
```
bernabe9: this looks like boilerplate code too

src/components/user/LoginForm.js
```diff
@@ -10,47 +10,53 @@ import {
...
+  >
+    {({ isSubmitting, status }) => (
+      <Form>
+        { status && <strong>{status}</strong> }
+        <div>
```
juanlacruz: I know that is not relevant on this pr, but as Felipe said regarding "style's" comment, I think we can remove these divs right?

src/components/user/LoginForm.js
juanlacruz: I think that initial values are needed, don't they? maybe you're suggesting another approach without adding this code here.

src/components/common/Input.js
nsantos16: I put setStatus that set status field in case of API response with an error but it's better to put all on errors field

src/components/user/LoginForm.js
nsantos16: Yep, even if your form is empty by default, you must initialize all fields with initial values otherwise React will throw an error saying that you have changed an input from uncontrolled to controlled.
What do you suggest Berna?

bernabe9: I suggest doing this: https://github.com/rootstrap/react-redux-base/pull/126#pullrequestreview-256358010 to automatically initialize the form with default empty values.

src/components/common/Input.js
nsantos16: I decided to use `setStatus` for backend errors for fixed this [issue with the backend response](https://github.com/jaredpalmer/formik/issues/150)

nsantos16:
@bernabe9 I added `<CustomFormik />` in order to improve reusability of submission functions, WDYT?

src/actions/sessionActions.js
```diff
@@ -1,27 +1,16 @@
...
-    try {
-      await sessionApi.logout();
-      sessionService.deleteSession();
-      sessionService.deleteUser();
-    } catch (err) {
-      throw (err);
-    }
+    await sessionApi.logout();
+    sessionService.deleteSession();
+    sessionService.deleteUser();
```
peter7z: What is the intention of the changes in these 3 lines?
Remember logout is not a 'form' action. This could potentially lead to unhandeled promise rejection errors.

PR #127, opened 2019-07-01
Add new useDispatch hook

peter7z:
Couldn't we use something like this? Since we will be doing it most time we use useDispatch...

### DIFF
--- a/src/hooks/useDispatch.js
+++ b/src/hooks/useDispatch.js
@@ -1,0 +1,10 @@
+import { useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+
+export default (action, ...dependencies) => {
+  const dispatch = useDispatch();
+  return useCallback(
+    payload => dispatch(action(payload)),
+    [dispatch, ...dependencies]
+  );
+};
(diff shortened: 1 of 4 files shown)


src/hooks/useDispatch.js
```diff
@@ -0,0 +1,10 @@
+import { useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+
+export default (action) => {
```
mcousillas6: Really nice idea! 👏 
But, if we are going to abstract this, we need to also forward the dependencies that will be watched on the callback
something like:
```javascript
export default (action, dependencies) => {
  const dispatch = useDispatch();
	  return useCallback(
	    user => dispatch(action(user)),
	    [dispatch, ...dependencies]
	  );
	};
```
Otherwise, if the action uses a value from the scope, the callback won't get updated propperly

src/hooks/useDispatch.js
```diff
@@ -0,0 +1,10 @@
...
+  const dispatch = useDispatch();
+  return useCallback(
+    user => dispatch(action(user)),
```
bernabe9: user?

PR #128, opened 2019-07-02
Prettier integration and refactor

nsantos16:
## Description

Based on  [PR](https://github.com/rootstrap/eye-level-flex-react/pull/5) made by @aguscha333 

## Pros
- Easy to adopt
- Helping Newcomers
- Building and enforcing a style guide

## Cons
- In some cases it's better unformatted code

## Documentation
- [Prettier docs](https://prettier.io/docs/en/install.html)
- [Precommit with prettier](https://prettier.io/docs/en/precommit.html)
- [Formatting Code Automatically](https://facebook.github.io/create-react-app/docs/setting-up-your-editor#formatting-code-automatically)
## Changes

- Added prettier
- Added hook to commits with husky
- Added lint-staged
- Configuration of `.eslintrc` and `.prettierrc`


.eslintrc
```diff
@@ -51,7 +55,7 @@
     "no-multi-spaces": 1,
     "block-spacing": 1,
     "brace-style": 1,
-    "comma-dangle": 0,
+    "comma-dangle": [1, "never"],
```
aguscha333: I would leave this line as it was before. It's a matter of tase really. In my PR to my project I changed it cause we decided we wanted to go that way.

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "printWidth": 100,
```
aguscha333: We should discuss what is the preferred `printWidth`, default is  `80`

aguscha333:
@nsantos16 to fix the code climate issues you need to specify the `.eslintrc` config path.

`.codeclimate.yml`
```
plugins:
  eslint:
    enabled: true
    config:
      config: ~/.eslintrc (add these 2 config lines)
```

PR #131, opened 2019-07-09
Updated some scripts

peter7z:
Just some script refactoring.
Read the code comments.
All scripts but deploy have been tested.


package.json
```diff
@@ -12,29 +12,30 @@
...
     "clean-dist": "npm run remove-dist && mkdir dist",
     "clean-ssr": "rimraf ./server/build",
     "remove-dist": "rimraf ./dist",
-    "prebuild": "npm run clean-dist && npm run lint && npm run test",
+    "prebuild": "npm-run-all clean-dist lint test",
     "build": "babel-node tools/build.js",
     "build:server": "babel-node server/scripts/buildServer.js",
     "build:client": "babel-node server/scripts/buildClient.js",
-    "test": "cypress run",
+    "test": "run-p -r start cypress-run",
```
peter7z: This is a race. `yarn start` will never end so the script will end when the tests finish. Previously you had to manually run the dev server before runing your tests. Just running `yarn build` would fail, for example.

package.json
```diff
@@ -12,29 +12,30 @@
...
+    "cypress-run": "cypress run",
     "test:open": "cypress open",
     "test:ci": "cypress run --key e83e86fd-9c47-4b2b-980d-517c9aff5974",
     "test:spec": "cypress run --spec",
     "jest": "jest",
-    "jest:cover": "npm run jest -- --coverage",
-    "jest:watch": "npm run jest -- --watch",
-    "deploy": "npm run build && babel-node tools/deployS3",
+    "jest:cover": "yarn jest -- --coverage",
```
peter7z: I don't know if yarn will run npm or yarn if this was `npm jest --...`
But since we are using yarn I just changed it for yarn.

PR #132, opened 2019-07-10
Upgrade to babel 7

manema:
Description:
- Stage presets(deprecated) have been replaced with individual proposal plugins.
- babel-polyfill has been replaced in favor of @babel/polyfill.

- Little fix to webpack config path in analyzeBundle.js file (this has nothing to do with the upgrade).

Upgrade documentation:
[https://babeljs.io/docs/en/v7-migration](url)


package.json
```diff
@@ -85,23 +85,39 @@
...
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "@babel/register": "^7.0.0",
     "@cypress/webpack-preprocessor": "4.0.3",
     "assets-webpack-plugin": "3.9.5",
     "autoprefixer": "9.1.1",
     "aws-sdk": "2.292.0",
-    "babel-cli": "6.26.0",
-    "babel-core": "6.26.3",
+    "babel-core": "^7.0.0-bridge.0",
```
bernabe9: do we still need this library?

.babelrc
```diff
@@ -1,22 +1,52 @@
...
+          {
+            "ssr": true
+          }
+        ]
       ]
     }
-  }
+  },
+  "plugins": [
+    "@babel/plugin-syntax-dynamic-import",
```
bernabe9: why so many plugins? do we need them all?

package.json
```diff
@@ -85,23 +85,39 @@
     "validate.js": "0.11.1"
   },
   "devDependencies": {
+    "@babel/cli": "^7.0.0",
```
bernabe9: use exact versions

webpack/webpack.config.client.js
```diff
@@ -6,7 +6,7 @@ import path from 'path';
 import Dotenv from 'dotenv-webpack';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import { GenerateSW } from 'workbox-webpack-plugin';
-import 'babel-polyfill';
+import '@babel/polyfill';
```
bernabe9: did you test the polyfill in IE?

Also, it looks like it has been deprecated in Babel 7.4.0: https://babeljs.io/docs/en/babel-polyfill

.babelrc
manema: Yes, stage presets have been deprecated in favor of just the plugins. Since we had the stage-1 installed, we need to install all the equivalent plugins from stage 1 to 3.

package.json
manema: Yes, because we have jest as a dependency. 
> use "babel-core": "^7.0.0-bridge-0" if jest or jest-cli is a dependency
[https://github.com/babel/babel-upgrade](url)

webpack/webpack.config.client.js
manema: No, you did it :grin:

package.json
manema: Done.

package.json
```diff
@@ -56,6 +56,7 @@
   "dependencies": {
     "classnames": "2.2.6",
     "connected-react-router": "6.4.0",
+    "core-js": "3.0.1",
```
bernabe9: I think this is a dev dependency, right?

manema: Yes, thank you for pointing it, It seems like I forgot add the --dev flag.

webpack/webpack.config.client.js
```diff
@@ -19,7 +20,7 @@ const GLOBALS = {
 export default {
   resolve,
   devtool: 'source-map',
-  entry: ['babel-polyfill', path.resolve(__dirname, '../server/client')],
+  entry: ['@babel/polyfill', path.resolve(__dirname, '../server/client')],
```
bernabe9: is `@babel/polyfill` still needed?

package.json
```diff
@@ -85,27 +85,43 @@
     "validate.js": "0.11.1"
   },
   "devDependencies": {
+    "@babel/cli": "7.5.0",
+    "@babel/core": "7.5.4",
+    "@babel/node": "7.5.0",
+    "@babel/plugin-proposal-class-properties": "7.5.0",
+    "@babel/plugin-proposal-decorators": "7.4.4",
+    "@babel/plugin-proposal-do-expressions": "7.5.0",
```
agustito37: are we using this plugin?

package.json
```diff
@@ -85,27 +85,43 @@
...
+    "@babel/plugin-proposal-export-default-from": "7.5.2",
+    "@babel/plugin-proposal-export-namespace-from": "7.5.2",
+    "@babel/plugin-proposal-function-sent": "7.5.0",
+    "@babel/plugin-proposal-json-strings": "7.2.0",
+    "@babel/plugin-proposal-logical-assignment-operators": "7.2.0",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "7.4.4",
```
agustito37: same

package.json
```diff
@@ -85,27 +85,43 @@
...
+    "@babel/core": "7.5.4",
+    "@babel/node": "7.5.0",
+    "@babel/plugin-proposal-class-properties": "7.5.0",
+    "@babel/plugin-proposal-decorators": "7.4.4",
+    "@babel/plugin-proposal-do-expressions": "7.5.0",
+    "@babel/plugin-proposal-export-default-from": "7.5.2",
+    "@babel/plugin-proposal-export-namespace-from": "7.5.2",
+    "@babel/plugin-proposal-function-sent": "7.5.0",
+    "@babel/plugin-proposal-json-strings": "7.2.0",
+    "@babel/plugin-proposal-logical-assignment-operators": "7.2.0",
```
agustito37: same

webpack/webpack.config.client.js
agustito37: babel-polyfill is deprecated and includes them, I think if you explicitly set core-js and runtime you don't need it.

manema: I knew @babel/polyfill was deprecated in favor of core-js and runtime, but as I haven't found any documentation/tutorial that specify if you have to keep the @babel/polyfill at the entry or take it away, I just leave it as it was.
Anyway we made a try with @bernabe9 removing it and seems work fine. So thanks for your contribution @agustito37

agustito37: Cool, I didn't find it too, so I wasn't sure.

package.json
manema: To be honest I don't know how babel exactly work behind the scenes, I agree with you that we aren't likely use those one, neither this @babel/plugin-proposal-pipeline-operator, but I'm not sure if removing it could cause any kind of problem in the future.

agustito37: Well those are proposals (stage 1 features), seems they aren't required by us to transpile. I don't know if someone added them for a reason or maybe he thought we could use them in the future, anyway it's a good moment to review and remove if we are not using them.

  * I only pointed out the ones I never saw in practice and seem unnecessary:
- https://github.com/tc39/proposal-logical-assignment
- https://github.com/tc39/proposal-nullish-coalescing
- https://github.com/tc39/proposal-do-expressions

  Maybe someone else can give us their opinion on this.

manema: Totally agree with you. Do you have any thought in this matter @fsarachu @mcousillas6 @aguscha333 @juanlacruz ? Or would you know about someone who may be of help?

mcousillas6: I think we can try removing them and testing, looks like we are not using them 🤷‍♂ .

mariosouto: I think that some of them are worth it, but since they are just proposals yet i'm not sure of including them on the base

PR #143, opened 2019-09-26
Preloading views

nsantos16:
## Description
Added `PreloadLink` for implement lazy loading on components, I saw this behavior on some Gatsby sites and I really liked how discrete and efficient it is.

## Pros

-  Code splitting (Less time of loading on the views)
- `Suspense`  and `lazy` don't require a lot of changes, as for example other libraries such as `react-loadable`
- Better user experience

## Cons

- Problems with SSR? @bernabe9 @santiestra 
- Tell me in the comments 😄

## Preview
Kapture 2019-09-26 at 10 23 58

## Documentation
- https://reactjs.org/docs/code-splitting.html
- https://medium.com/@rossbulat/react-lazy-suspense-and-concorrent-react-breakdown-with-examples-2758de98cb1c
- https://blog.maximeheckel.com/posts/preloading-views-with-react

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, Suspense } from 'react';
 import { ConnectedRouter } from 'connected-react-router';
 import { Switch } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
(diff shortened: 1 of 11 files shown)


agustito37:
Should we lazy preload all the routes as default or this is just an example?

src/components/App.js
```diff
@@ -27,11 +27,13 @@ const App = () => {
         <Header />
         <ConnectedRouter history={history}>
           {checked && (
-            <Switch>
-              {routes.map((route, index) => (
-                <RouteFromPath key={`route${index}`} {...route} authenticated={authenticated} />
-              ))}
-            </Switch>
+            <Suspense fallback="Loading">
```
bernabe9: Do we want to use the same fallback for all the pages? 🤔 Is it possible to don't use any fallback component?

agustito37: can we dispatch an action instead maybe?

PR #147, opened 2019-10-22
Redux persist

ConstanzaCurzio:
### Tasks
- Adds redux-persist
- Uses localforage to save data
- Deletes redux-react-session

### Reasons
Nowadays almost every project needs to persist custom variables in the browser, so it's a good asset to have in the base project

redux-persist let's you persist reducers, more specifically, the keys of the reducers you want. So it's totally fine if you want to save the token key of the User reducer, but not the error key.

This implementation is saving data in localforage, but there are many other alternatives like [redux-persist-cookie-storage](https://github.com/abersager/redux-persist-cookie-storage)

I removed redux-react-session since it was redundant to have two ways of storage.

Easier control for session dispatchs. With redux-react-session actions were dispatched unexpectedly in some cases.

### Future work
- Status reducer


src/reducers/createReducer.js
```diff
@@ -2,5 +2,5 @@ import produce from 'immer';
 
 export default (initialState, actionHandlers) => (state = initialState, action) =>
   produce(state, draft =>
-    actionHandlers[action.type] ? actionHandlers[action.type](draft, action.payload) : state
+    actionHandlers[action.type] ? actionHandlers[action.type](draft, action) : state
```
peter7z: Why?

ConstanzaCurzio: This shouldn't depend on the action to have a payload, it's really coupled.

peter7z: Why do you say it depends on it having a payload? If it doesn't it will just evaluate to `undefined`, whats the problem with that?

peter7z: https://github.com/rootstrap/react-redux-base/blob/master/src/actions/createAction.js

ConstanzaCurzio: Because it does `action.payload`, this shouldn't know the implementation or that action has a payload.

mcousillas6: @peter7z  the problem of just passing `.payload` here is that if your action does not uses payload as it's root key to pass information to the reducer, there's no way to get it on you handlers. For example:
````javascript
const loginSuccess = user => ({
  type: actionTypes.LOGIN_SUCCESS,
  user,
});

const handleUpdateUser = (state, { user }) => {
  state.user = user;
};
```

With the previous createReducer, that wouldn't work. And it's not a good design choice to limit your action's keys to just payload just because we have an optional helper to create actions.
Moreover, the react-native base had this exact implementation, so i'm not sure why the web one had the `.payload` in the first place 🤷‍♂

peter7z: Our action creators only allow adding data to the action under `payload`.
I think that if we decide that we don't like this system then we should just remove `createAction`.

src/reducers/index.js
```diff
@@ -1,12 +1,23 @@
...
 import { reducer as form } from 'redux-form';
-import { sessionReducer as session } from 'redux-react-session';
 import { connectRouter } from 'connected-react-router';
+import localForage from 'localforage';
+import { persistReducer } from 'redux-persist';
+import autoMergeLevel2 from 'redux-persist/lib/stateReconciler/autoMergeLevel2';
+
+import session from './sessionReducer';
+
+const sessionPersistConfig = {
```
agustito37: I don't this should be here.

ConstanzaCurzio: Why not? 🤔

src/reducers/createReducer.js
ConstanzaCurzio: Generally, it's good if change doesn't impact everywhere. So if tomorrow we change `createAction`, ideally, it shouldn't impact on something like `createReducer`. That is more likely to introduce bugs

mcousillas6: The thing is that you are tying up the whole reducer structure to that decision, so, for example, if you are wiring up a library that exposes it's own actions, our reducers would not see the associated information.

About the `I think that if we decide that we don't like this system then we should just remove createAction.` I think it's good to have a standard flow in place, but there's no need to tie the whole implementation to it just for the sake of it. This change does not impact the way you use `createAction`.

peter7z: Ok, makes sense.

agustito37: Agree with @peter7z if you send the action and not the payload, the reducers become inconsistent with the actions; and you are adding an unnecessary deconstructing extra step.

src/reducers/index.js
agustito37: This is part of the app's configuration, a higher abstraction than the reducer, it involves reducers and local persistence. I would define this on the `app.js` and pass this conf as an argument.

src/reducers/createReducer.js
agustito37: Mmm I didn't read the previous message, it has sense. Maybe we can change the createActions but I'm not sure is worth.

src/reducers/index.js
ConstanzaCurzio: It made sense to define it in the reducer since it involve the reducer, but I get what you mean. Mmm the thing is we have 2 configureStore in web https://github.com/rootstrap/react-redux-base/tree/master/src/store so I didn't want to duplicate the config in each file.

agustito37: Ok, follow your heart. It's fine anyway.

PR #150, opened 2019-10-25
Status reducer

ConstanzaCurzio:
## Status Reducer 

Following [this](https://github.com/rootstrap/react-native-base/pull/74) PR in the RN base project

### Tasks 

Added a generic reducer that track the status of every action dispatched. The main goal is to reduce boilerplate and standardise status and error handling in one place.

The status reducer matches against ACTION_(REQUEST|ERROR|SUCCESS|RESET) to define the status of the action between STARTED|NOT_STARTED|LOADING|ERROR.
If the action has an error key, it will be stored alongside the status.

Added hooks to query the store.

This way, we only need to dispatch the correct action types, and we are good to go, the statusReducer takes care of the rest, and you can use the provided hooks useLoading useError and useStatus to query that state for each action.

Used it to show loading in Login and Signup form, and to handle error in the Login form.

### Future work
- Remove redux-forms

PS: thanks @mcousillas6 😁

### DIFF
--- a/src/components/user/LoginForm.js
+++ b/src/components/user/LoginForm.js
@@ -1,12 +1,12 @@
 import React, { memo } from 'react';
-import { func, string } from 'prop-types';
+import { func } from 'prop-types';
 import { Field, reduxForm } from 'redux-form';
 import { useIntl, defineMessages, FormattedMessage } from 'react-intl';
 
 import Loading from 'components/common/Loading';
 import Input from 'components/common/Input';
 import { validations, login } from 'utils/constraints';
-import { LOADING } from 'constants/status';
+import { LOADING, ERROR } from 'constants/status';
 import { useStatus } from 'hooks';
 import { LOGIN } from 'actions/actionTypes';
 
@@ -15,13 +15,13 @@ const messages = defineMessages({
   password: { id: 'login.form.password' }
 });
 
-export const LoginForm = ({ handleSubmit, error }) => {
+export const LoginForm = ({ handleSubmit }) => {
   const intl = useIntl();
-  const { status } = useStatus(LOGIN);
+  const { status, error } = useStatus(LOGIN);
 
   return (
     <form onSubmit={handleSubmit}>
-      {error && <strong>{error}</strong>}
+      {status === ERROR && <strong>{error}</strong>}
       <div>
         <Field
           name="email"
@@ -47,8 +47,7 @@ export const LoginForm = ({ handleSubmit, error }) => {
 };
 
 LoginForm.propTypes = {
-  handleSubmit: func.isRequired,
-  error: string
+  handleSubmit: func.isRequired
 };
 
 export default reduxForm({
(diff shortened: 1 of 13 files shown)


src/reducers/statusReducer.js
```diff
@@ -0,0 +1,38 @@
+import { produce } from 'immer';
+import { NOT_STARTED, LOADING, SUCCESS, ERROR } from 'constants/status';
+
+const handleAction = (state, action) => {
+  const { type, payload } = action;
+
+  const matchesStart = /(.*)_(REQUEST)/.exec(type);
```
peter7z:
```suggestion
  const matchesStart = /.*_(REQUEST)/.exec(type);
```

src/reducers/statusReducer.js
```diff
@@ -0,0 +1,38 @@
...
+  const matchesError = /(.*)_(ERROR)/.exec(type);
+  const matchesReset = /(.*)_(RESET)/.exec(type);
+  const matchesSuccess = /(.*)_(SUCCESS)/.exec(type);
+
+  let status = NOT_STARTED;
+  let key = null;
+
+  if (matchesStart) {
+    const [, requestName] = matchesStart;
```
peter7z:
```suggestion
    const [requestName] = matchesStart;
```

mcousillas6: This would break the array destructuring I think, since we want the second item

src/components/user/LoginForm.js
```diff
@@ -36,15 +41,13 @@ export const LoginForm = ({ handleSubmit, error, submitting }) => {
       <button type="submit">
         <FormattedMessage id="login.form.submit" />
       </button>
-      {submitting && <Loading />}
+      {status === LOADING && <Loading />}
```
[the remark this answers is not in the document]
mcousillas6: If you check both, it's easier to `useStatus` and match the enum like coni is doing in this case, otherwise `useLoading`

src/constants/status.js
```diff
@@ -0,0 +1,7 @@
+export const NOT_STARTED = 'NOT_STARTED';
```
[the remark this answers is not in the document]
mcousillas6: I think this should stay singular, since we are using it as an enum, wdyt?

src/components/user/SignUpForm.js
```diff
@@ -44,14 +49,13 @@ export const SignUpForm = ({ handleSubmit, submitting }) => {
       <button type="submit">
         <FormattedMessage id="login.form.submit" />
       </button>
-      {submitting && <Loading />}
+      {status === LOADING && <Loading />}
```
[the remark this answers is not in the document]
ConstanzaCurzio: Yes, I'll change it

src/reducers/statusReducer.js
peter7z: Yeaa, my bad, I thought you wanted to capture the `REQUEST` part.

peter7z: What I think you will be able to do is remove parenthesis around `REQUEST` etc...

ConstanzaCurzio: Yeah, that's right

PR #153, opened 2019-11-05
Switch redux forms for a custom form hook

peter7z:
Replaces redux-form with the useForm hooks from react-native-base.
Just some small adjustments:
- `onChangeText` had to be converted to `onChange` and the event object had to be deconstructed.
- Used `{ fullMessages: false }` as default option on the validation hook, since that was what we were using.

This PR made me think think that we should really get our error handling together...
For more info check: https://github.com/rootstrap/react-native-base/pull/86

### DIFF
--- a/src/components/user/LoginForm.js
+++ b/src/components/user/LoginForm.js
@@ -1,13 +1,12 @@
 import React, { memo } from 'react';
 import { func } from 'prop-types';
-import { Field, reduxForm } from 'redux-form';
 import { useIntl, defineMessages, FormattedMessage } from 'react-intl';
 
 import Loading from 'components/common/Loading';
 import Input from 'components/common/Input';
-import { validations, login } from 'utils/constraints';
+import { login } from 'utils/constraints';
 import { LOADING, ERROR } from 'constants/status';
-import { useStatus } from 'hooks';
+import { useStatus, useForm, useValidation, useTextInputProps } from 'hooks';
 import { LOGIN } from 'actions/actionTypes';
 
 const messages = defineMessages({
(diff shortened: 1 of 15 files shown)


src/hooks/useForm.js
```diff
@@ -0,0 +1,87 @@
...
+  const [blurredFields, setBlurredFields] = useState({});
+
+  const handleSubmit = useCallback(
+    event => {
+      event.preventDefault();
+      const newErrors = validator(values) || {};
+      const valid = !Object.values(newErrors)
+        .filter(error => !!error)
+        .reduce((acc, error) => {
+          error.forEach(e => acc.push(e));
```
agustito37: why are you creating another array? this reduce seems redundant.

agustito37:
```suggestion
          acc.push(..error);
```

src/hooks/index.js
```diff
@@ -2,3 +2,6 @@ export { default as useSession } from './useSession';
 export { default as useDispatch } from './useDispatch';
 export { default as useLoading } from './useLoading';
 export { default as useStatus } from './useStatus';
+export { default as useForm } from './useForm';
```
agustito37: This entire file seems redundant.

mcousillas6: Yep, there's no point on having an index.js if you are always going to import single files

src/hooks/useForm.js
mcousillas6: Check IE support for this. I had to do it this way on RN because of hermes method availability, but it would look much cleaner

src/hooks/index.js
peter7z: For me is much cleaner to have all this imports on the same line vs having each in a separate one.

src/hooks/useForm.js
peter7z: Flat is nos supported on IE nor Edge... But i think Babel will transpile this to compatible js?

src/components/user/LoginForm.js
```diff
@@ -1,41 +1,60 @@
...
+
+export const LoginForm = ({ onSubmit }) => {
   const intl = useIntl();
   const { status, error } = useStatus(LOGIN);
+  const validator = useValidation(login);
+  const { values, errors, handleValueChange, handleSubmit, handleBlur } = useForm(
+    {
+      initialValues: {
+        email: '',
+        password: ''
```
ConstanzaCurzio: Can we avoid defining this? Seems a little redundant

peter7z: This sets a value when the form is initialized. If we don't have an initial value, once a value is set, we get a warning that a field changed from uncontrolled to controlled. Maybe we could implement something so that this wouldn't be necessary. Any ideas?

src/hooks/index.js
mcousillas6: It's personal preference, and it's not a bad practice to define the api of a module using index.js, so keep it

agustito37: It's fine by me. Thanks for the clarification.

src/hooks/useForm.js
agustito37: I think too, not sure given is a new proposal, but you can try.

src/hooks/useTextInputProps.js
```diff
@@ -0,0 +1,12 @@
+import { useCallback } from 'react';
+
+export default ({ handleValueChange, handleBlur, values, errors }) =>
+  useCallback(
+    fieldKey => ({
+      value: values[fieldKey],
```
ConstanzaCurzio: Maybe here you can put `value: values[fieldKey] || '',` or something like that @peter7z so you can remove the initial values (my previous comment) 🤔

peter7z: Great, I think that will do it. I'll try and let you know. @ConstanzaCurzio

peter7z: Done! It's working, also added a default object for initialValues on useForm, I think we should add this on mobile!

PR #154, opened 2019-11-07
Version bumps

ConstanzaCurzio:


peter7z:
Great! What about adding the webpack package uodtawd in #152 ?

PR #156, opened 2019-11-11
Suspense wip

peter7z:
Hey guys.
I'm experimenting with the new [suspense for data fetching](https://reactjs.org/docs/concurrent-mode-suspense.html) react mechanic.

This is veeery early testing, this is not even present on react current builds. Still, I think it allows managing the loading and asynchronous aspect of data fetching in a very declarative and clean way.

As of right now all the logic is in a custom hook: `useResource`. It still lacks quite a bit of customizaiton ability
What do you guys think? Is it worth exploring a little bit more? Maybe try to move some of the logic into a redux middleware?
Please, tell me what you think!

For the lazy, this is what the component you see in the gif looks like:
```js
const HomePage = () => (
  <div>
    <Suspense fallback={<h3>Loading...</h3>}>
      <UserName />
    </Suspense>
  </div>
);
```

and this is the UserName component:
```js
const UserName = () => {
  const user = useResource(userResource);
  return <h2>{user}</h2>;
};
```

**Edit**: I've now moved some logic into a redux middleware, I think it's a little bit better now. Still a lot of room for improvement.

Gif :D
syspense


src/hooks/useResource.js
```diff
@@ -0,0 +1,33 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const resource = useSelector(selector);
+  const { error: responseError, status } = useStatus(actionName);
+
+  console.log('useResource', status, error);
+  if (status === SUCCESS) return resource;
+  if (status === ERROR) throw responseError;
+  if (status === LOADING) throw promise.current;
```
peter7z: This last lines are basically the contract we have with react to let it know when the resource is ready.
Return for a ready resource, throwing a promise to let it know we are still loading,
and throw an error if the fetching failed.

src/hooks/useResource.js
```diff
@@ -0,0 +1,22 @@
...
+    dispatch(resource);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const data = useSelector(selector);
+  const { error, status, promise } = useStatus(resourceName);
+
+  if (status === SUCCESS) return data;
+  if (status === ERROR) throw error;
+  if (status === LOADING) throw promise;
```
peter7z: This last lines are basically the contract we have with react to let it know when the resource is ready.
Return for a ready resource, throwing a promise to let it know we are still loading,
and throw an error if the fetching failed.

src/actions/createResource.js
```diff
@@ -0,0 +1,18 @@
...
+  const request = createAction(`${resourceName}_REQUEST`);
+  const type = createAction(`${resourceName}_RESOURCE`);
+
+  return {
+    resourceName,
+    selector,
+    success,
+    error,
+    request,
+    handler,
```
mcousillas6: thinking about naming, not sure if handler is the best for this since it will always be data-fetching related, right?

src/hooks/useResource.js
```diff
@@ -0,0 +1,22 @@
...
+import { LOADING, SUCCESS, ERROR } from 'constants/status';
+
+export default resource => {
+  const dispatch = useDispatch();
+  const { selector, resourceName } = resource;
+
+  useEffect(() => {
+    dispatch(resource);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
mcousillas6: what happens if you put the full deps here? 🤔

src/middleware/resources.js
```diff
@@ -0,0 +1,16 @@
+const isResouce = action => /_RESOURCE$/.test(action);
+
+export default ({ dispatch }) => next => async action => {
+  const { type, handler, success, error, request } = action;
+  if (!isResouce(type)) return next(action);
```
mcousillas6: do we only want to call handlers for resource types? 
Also, is it really necessary to have a type object to match, or can we just check if handler is a function (and suppose it's a resource, since it's following the defined behaviour of dispatching `handler, success, error and request`).

we could check `typeof handler`, similar to what thunks does. wdyt?

src/components/user/UserName.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import { userResource } from 'actions/testActions';
+import { useResource } from 'hooks';
+
+export default function UserName() {
+  const user = useResource(userResource);
+  return <h2>{user}</h2>;
```
mcousillas6: Question about patterns, WDYT about wrapping suspense at the same level that you are fetching the resource? 
That way you could have specific loading states for your asynchronous components, and you don't have to wait for the request to finish to start rendering the ui that does not depend on the userResource.

src/hooks/useResource.js
```diff
@@ -0,0 +1,22 @@
...
+import { useStatus } from 'hooks';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { LOADING, SUCCESS, ERROR } from 'constants/status';
+
+export default resource => {
+  const dispatch = useDispatch();
+  const { selector, resourceName } = resource;
+
+  useEffect(() => {
```
mcousillas6: fetching on a useEffect while using suspense is not the best way. The idea of using suspense is fetch as you render, so calling your data fetch (in this case, the action dispatch) earlier, before the component that has suspense starts rendering.

https://es.reactjs.org/docs/concurrent-mode-suspense.html#start-fetching-early

From what've read on the docs, the best approach is to fetch on event handlers (so, onPress for example), and in some cases, fetch at the top level of your file, before render is even executed:
(Also from the docs)
```javascript
const initialResource = fetchProfileData(0);

function App() {
  const [resource, setResource] = useState(initialResource);
....}
```

src/hooks/useResource.js
mcousillas6: Hate having to throw as a pattern, but for now it seems that's the only way the api works 😞

src/middleware/resources.js
peter7z: The thing is, every action needs to have a type. (At least redux logger will give us a warning if it doesn't)
If we could somehow disable that warning... Still, I think it's nice to be able to see that this resource action was dispatched on redux dev tools.
Regarding the handler thing, if we wanted we could always fetch if the action has a handler method (I think this could reduce a lot of the boilerplate we now have on action files)
I think this (even separated from suspense) is a good discussion topic.

src/components/user/UserName.js
peter7z: We can't wrap in JSX because then the throws would fall outside `<Suspense>`.
But yes, we could use a HOC or something.

src/hooks/useResource.js
peter7z: Good catch! Didn't notice this. I'll think of a way of implementing this.

peter7z: What about dispatching the resource to start the fetch, and consuming the data with `useResource`

PR #157, opened 2019-11-13
Add thunk action creators and middleware that handle them

peter7z:
Just an idea...

This will provide a creator that will let you create your actions like this:
```js
export const login = createActionWithThunk('LOGIN', async user => {
  try {
    const {
      data: { user: createdUser }
    } = await userService.login({ user });
    return createdUser;
  } catch (err) {
    throw err.data.error;
  }
});
```

I this example the creator will return a dispatchable action with the name `LOGIN_REQUEST`.
If this action is dispatched it will be caught by the middleware and it will execute the thunk. If no error is thrown it will automatically dispatch a success action with the result as payload, otherwise it will dispatch the error action with the error as payload.

This object also has 2 properties which you can destrucure as follows:

```js
export const {
  success: loginSuccess,
  error: loginError,
} = login;
```

Theese 2 (and the object itself too) can be used directly as the action names on the reducer, like so:

```js
import { loginSuccess, loginError } from 'actions/userActions';

const actionHandlers = {
  [loginSuccess]: (state, { payload }) => {
    state.user = payload;
  },
  [loginError]: (state, { payload }) => {
    state.error = payload;
  },
};
```

I think this can reduce a lot of boilerplate for our actions, since we always have the 3 actions and always dispatch success or failure depending on the promise result.

I also think that it can add complexity to how the base works, I don't know... I like it.

What do you think?

**Edit:** Inline doc and readme section added! Preview:

doc


src/actions/createThunk.js
```diff
@@ -0,0 +1,24 @@
+import createAction from './createAction';
+
+const SUCCESS = 'SUCCESS';
+const ERROR = 'ERROR';
+const REQUEST = 'REQUEST';
```
mcousillas6: you can already import these from` constants/status`

peter7z: There is no REQUEST in status constants, shall I add it?

src/actions/createThunk.js
```diff
@@ -0,0 +1,37 @@
...
+ * );
+ * export const { success, error } = getProfile;
+ */
+
+export default (actionName, thunk) => {
+  const request = createAction(`${actionName}_${REQUEST}`);
+  const error = createAction(`${actionName}_${ERROR}`);
+  const success = createAction(`${actionName}_${SUCCESS}`);
+
+  const action = (...params) => ({
```
mcousillas6: i'd add request as part of the return here as well, sometimes you want to add handlers there (to clear the old result for example)

src/middleware/actionThunks.js
```diff
@@ -0,0 +1,13 @@
+export default ({ dispatch }) => next => async action => {
+  next(action);
+
+  const { thunk, success, error } = action;
+  if (typeof thunk === 'function') {
+    try {
+      const response = await thunk();
+      dispatch(success(response));
+    } catch (err) {
+      dispatch(error(err));
```
mcousillas6: just to double check, are we sure that dispatching error like this leaves the `error` key on the status reducer with this value? or do we have to do `error.payload` aftwerwards?

peter7z: I'm not sure I understand what you are asking.
Status reducer will assign the payload (`err` in this case) to the error corresponding key.
`err` will be whatever error is thrown inside the thunk.

docs/redux.md
```diff
@@ -0,0 +1,72 @@
...
+To access this information on a component the `useStatus` hook is provided.
+
+Here is a simple example:
+
+```js
+import { useStatus, useDispatch } from 'hooks';
+import { getProfile } from 'src/actions/userActions';
+import { SUCCESS, LOADING } from 'constants/status';
+
+const MyComponent = () => {
```
manuelcheveste: I'm still not exactly sure why in some places we use arrow function instead of the vanilla one when we don't make use of the implicit return.
In my case I would prefer to use `function MyComponent() { }`.
Is just an opinion btw :)

package.json
```diff
@@ -88,7 +88,6 @@
     "redux": "4.0.4",
     "redux-logger": "3.0.6",
     "redux-persist": "6.0.0",
-    "redux-thunk": "2.3.0",
```
manuelcheveste: I probably will prefer to keep redux thunk, maybe you have to do something before using the middleware.

peter7z: I think the vast majority of cases will just use the custom middleware, if you think you need redux-thunk you can just install it.

docs/redux.md
peter7z: I think we all just got used to arrow functions?

mcousillas6: We use arrows by default, I think it helps to avoid some common mistakes like tracking `this` and not needing to bind stuff in most cases, but it's a matter of personal preference. 
I did some research a few months ago, and there's no notable improvement (performance-wise) between using one or the other.

manuelcheveste: Usually you don't have to worry about binding and this on `function ()` and I think is important that a js dev should know about how it this works. Either way, in the case you don't want scope, you can use arrow func, but most of the cases, you will receive everything you need in a component by the props or hooks.

I didn't really worry about performance. But what I like of `function` is this neat export like `export default functionMyComponent() { }`.
You can't do this with arrow functions

mcousillas6: Yep you're right but normally you'd have to declare at least propTypes for your component, so the export default is not a really common use-case. 
Having said that, again, I think both work fine and it's personal preference, in this case, i'd keep arrows because the whole base is done that way and we should keep consistency

manuelcheveste: You can still use propTypes. I know consistency is important, but both work like functions, so there is not really a difference between. I still would use arrow functions on inline functions. I just don't like to enforce this types of consistence. I mean, there are bigger things like, yeah you don't go arrow defining reducers in a component. But this is just a feature from the language

mcousillas6: Yep, definitely agree on that, but features should be used with a purpose, not just use it for the sake of it 😉 sometimes its better to not use something to maintain other quality attributes (maintainability being a really big one)

PR #161, opened 2019-11-21
Use shallow equal

ConstanzaCurzio:
### Description 
This was causing unnecessary renders. We update the session in every authenticated request, since almost always the token we save from the request is the same we already have stored, we can check this before updating it. We can use shallow equal in useSession so it compares on object keys (not deep) and avoid re-computing and re-rendering the component.

### DIFF
--- a/src/hooks/useSession.js
+++ b/src/hooks/useSession.js
@@ -2,9 +2,9 @@ import { useSelector, shallowEqual } from 'react-redux';
 
 const useSession = () =>
   useSelector(
-    ({ session }) => ({
-      authenticated: session.authenticated,
-      user: session.user
+    ({ session: { authenticated, user } }) => ({
+      authenticated,
+      user
     }),
     shallowEqual
   );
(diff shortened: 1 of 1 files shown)


peter7z:
Isn't reference the default comparison?

src/hooks/useSession.js
```diff
@@ -1,9 +1,12 @@
-import { useSelector } from 'react-redux';
+import { useSelector, shallowEqual } from 'react-redux';
 
 const useSession = () =>
-  useSelector(({ session }) => ({
-    authenticated: session.authenticated,
-    user: session.user
-  }));
+  useSelector(
+    ({ session }) => ({
```
kevquincke: is it possible to do something like:
```suggestion
    ({ session: { authenticated, user } }) => ({
        authenticated,
        user
    }), ...
```
?

manuelcheveste: This looks like nesting if statements, I would prefer not to touch it

ConstanzaCurzio: @kevquincke good idea! I'll do it now

@manuelcheveste I didn't get what you said 😅 it's just destructuring the object

manuelcheveste: tbh, is not that bad like this, but It can become easy to abuse this feature, and then you can end with something like this
```
session: {
  user: { document: { type } }
}
```
and this can go on and on and it will become hard to understand

PR #163, opened 2019-12-05
Move linters to Code Climate

mariosouto:
We discussed this on the Code Quality group that it would be better to centralize code quality checks like linters and other tools in Code Climate like we are doing with the other technologies (Rails, iOS and Android)


bernabe9:
@mariosouto can you provide a description of why we are doing this change?

PR #169, opened 2020-01-10
Fix log out when unauthorized. Add .env.sample

### LINKED ISSUE #168: Log out is not working when api returns 401 Unauthorized
### Problem

Nothing is done when api returns `Error 401 unauthorized`. 

It is expected in `applyDefaultInterceptors` in `defaultResponseInterceptors` to logout when this error occurs but it is not working:

```
const defaultResponseInterceptors = store => [
  async response => {
    if (response.ok) {
        ...
      }
    }
    if (response.status === UNAUTHORIZED) {
      store.dispatch(logout());
    }
    return response;
  }
];

export default (store, apiService) => {
  defaultResponseInterceptors(store).forEach(interceptor =>
    apiService.responseInterceptors.use(interceptor)
  );
};
```
The problem in the  previous code is that in `apiService.responseInterceptors.use(interceptor)` the use method expects two functions as parameters (first one in case of fulfill and second one in case of failure) but is only being passed the fulfill parameter. 

### Possible Solution

A possible solution we encountered was to divide de `defaultResponseInterceptors` in two separate functions: one for fulfill and another one in case of failure (with the unauthorized logic in it) to pass both of them to the `use` method. 

```
apiService.responseInterceptors.use(
    defaultFulfillResponseInterceptor(store),
    defaultFailureResponseInterceptor(store)
  );
```

lucia-santos:
Fix #168


bernabe9:
@lucia-santos Can you change the description to automatically close the issue after merging?

.env.sample
```diff
@@ -0,0 +1,6 @@
+API_URL=https://rails5-api-base.herokuapp.com/api/v1
+RELEASE_STORE_FILE=
+RELEASE_KEY_ALIAS=
+RELEASE_STORE_PASSWORD=
+RELEASE_KEY_PASSWORD=
+AMPLITUDE_API_KEY=
```
nsantos16: I think there's no integration with Amplitude, also in the readme said `env.example` and contain theses keys
```
API_URL=http://your-api-url.com
CABLE_URL=wss://your-api-url.com/cable
AWS_BUCKET=bucket
AWS_REGION=region
AWS_ACCESS_KEY_ID=key_id
AWS_SECRET_ACCESS_KEY=secret_key
```

src/api/utils/applyDefaultInterceptors.js
```diff
@@ -26,33 +26,36 @@ const defaultRequestInterceptors = store => [
...
-      const token = headers.get(ACCESS_TOKEN);
-      if (token) {
-        const session = {
-          token,
-          uid: headers.get(UID),
-          client: headers.get(CLIENT)
-        };
-        store.dispatch(updateSession(session));
-      }
+const defaultFulfillResponseInterceptor = store => async response => {
```
mato2593: This is just a personal preference, but I prefer to have the `async response => {` in a separate line. I don't like two arrows in the same line 😕 .

.env.sample
bernabe9: Also, can we remove the `CABLE_URL` variable from README and `.env.sample`?

src/api/utils/applyDefaultInterceptors.js
bernabe9: I think `prettier` is changing this, @lucia-santos correct me if I'm wrong.

lucia-santos: Yep, you are right Berna. 
I like it better that way too mathi haha, i can change it

mato2593: 🤙 . I'm starting to notice that I usually differ with prettier's opinions 😅

Issue #170, opened 2020-01-22
S3 deploy script won't upload folders.


peter7z:
Our current S3 "syncDir" method will only sync the files on the directory, not other directories.

agustito37:
I can do this.

agustito37:
Solved on https://github.com/rootstrap/react-redux-base/pull/171

Issue #151, opened 2019-10-29
Make build Gzipping optional


fsarachu:
Currently the production build script is always gzipping all files, which is not required for all cases. Optimal setup for a static deployment would be delegating gzip to CloudFront or any server being used.

bernabe9:
Nicee. I would add this task to a production checklist in the README.

agustito37:
Solved on https://github.com/rootstrap/react-redux-base/pull/171

PR #171, opened 2020-01-22
Refactor s3 script to accept folders

agustito37:
- Now deployment to s3 accepts folders
- Use of async/await notation
- Added static website hosting URL for the bucket
- Removed compression by default, now you have an environment variable `GZIP_ENABLED=false` by default

Fix https://github.com/rootstrap/react-redux-base/issues/151
Fix https://github.com/rootstrap/react-redux-base/issues/170

### DIFF
--- a/tools/s3.js
+++ b/tools/s3.js
@@ -2,9 +2,11 @@ import S3 from 'aws-sdk/clients/s3';
 import fs from 'fs';
 import path from 'path';
 import mime from 'mime';
+import { consoleQuestion } from './helpers';
 
 class s3 {
-  constructor() {
+  constructor(distFolder) {
+    this.distPath = path.join(__dirname, distFolder);
     this.client = new S3({
       signatureVersion: 'v4',
       accessKeyId: process.env.AWS_ACCESS_KEY_ID,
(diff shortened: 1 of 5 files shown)


tools/s3.js
```diff
@@ -14,76 +17,64 @@ class s3 {
...
-          return;
-        }
+  async clearBucket(bucketObjects) {
+    try {  
+      for (let x = 0; x < bucketObjects.length; x += 1) {
+        const { Key: key } = bucketObjects[x];
+        const deleteParams = { Bucket: this.bucket, Key: key };
+        console.log(chalkProcessing(`Deleting: ${key}`));
+        await this.client.deleteObject(deleteParams).promise();
+      }
```
bernabe9: Is each iteration waiting for the promise?

tools/s3.js
```diff
@@ -14,76 +17,64 @@ class s3 {
...
-      const folderPath = path.join(__dirname, distPath);
-
-      fs.readdir(folderPath, (err, files) => {
-        if(!files || files.length === 0) {
-          reject(`provided folder '${distPath}' is empty or does not exist.`);
-          return;
-        }
+  async clearBucket(bucketObjects) {
+    try {  
+      for (let x = 0; x < bucketObjects.length; x += 1) {
```
bernabe9:
```suggestion
      const promises = bucketObjects.map(({ Key })=>
        this.client.deleteObject({ Bucket: this.bucket, Key };).promise();
      );
      await Promises.all(promises)
```

tools/s3.js
```diff
@@ -14,76 +17,64 @@ class s3 {
...
+      throw new Error(`error uploading bucket objects ${err}`);
+    }
+  }
 
-          fs.readFile(filePath, (error, fileContent) => {
-            if (error) { reject(error); }
+  async walkSync (currentDirPath) {
+    const objects = fs.readdirSync(currentDirPath);
+    for(let x = 0; x < objects.length; x += 1) {
+      await this.walkFileOrDir(currentDirPath, objects[x]);
```
bernabe9: Is this performing sequentially?

bernabe9:
Nice job! It would be great to log the bucket's URL for quick access to the deployed web app.

tools/s3.js
agustito37: yes, but I will remove this if there's no lower quotas.

tools/s3.js
agustito37: Cool, thanks!

tools/s3.js
agustito37: Yes, I'll change this.

tools/s3.js
agustito37: Indeed it is performing sequentially, but I will change this, I want to check AWS quotas before doing the change.

agustito37:
> Nice job! It would be great to log the bucket's URL for quick access to the deployed web app.

I tried that, seems the js is gzipped (it doesn't have the extension) so the deployed app doesn't work. I'll check this.

PR #178, opened 2020-03-04
Fix gzip support on s3 script

agustito37:
I tested locally and it's working.


agustito37:
I tested locally without the ContentEncoding and worked, I will maintain the additional parameter just in case, maybe someone can test it too.

PR #180, opened 2020-03-09
Rename succes actions

peter7z:
Don't you think this looks better? It's fewer lines of code too!

I think total coverage went down because I removed 4 lines of code that where being tested. Thats why CI is failing.


nsantos16:
I think this is the best option, because in the long term if you adopt the other alternative you gonna have a long list of successful actions in our actions files:
```
export const { success: loginSuccess } = login;	
export const { success: signUpSuccess } = signUp;	
export const { success: logoutSuccess } = logout;
...
```
Good job 💪

PR #182, opened 2020-03-11
Fix server error handling

martinjaimem:
When one promise fails, then the error isn't being catched

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -38,69 +38,65 @@ server
       }
     })
   )
-  .get('*', async (req, res) => {
-    try {
-      const { store } = configureStore(undefined, true);
-      const context = {};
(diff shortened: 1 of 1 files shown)


server/server.js
```diff
@@ -58,7 +58,7 @@ server
       const userLocale = getLanguageFromHeader(req.get('Accept-Language'));
       const messages = locales[userLocale];
 
-      Promise.all(promises).then(async () => {
+      await Promise.all(promises).then(async () => {
```
peter7z: Why is this function async?

santiestra: Good call, it has no await inside 🤔

peter7z: @martinjaimem Mind addressing this?

PR #189, opened 2020-03-16
Add dotenv plugin to prod webpack config

peter7z:
Production webpack configuration as missing dotenv plugin.
Thanks @acollazomayer for finding this issue.

### DIFF
--- a/webpack/webpack.config.prod.js
+++ b/webpack/webpack.config.prod.js
@@ -1,6 +1,7 @@
 import webpack from 'webpack';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin';
+import Dotenv from 'dotenv-webpack';
 import WebpackMd5Hash from 'webpack-md5-hash';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


bernabe9:
Was this change intentional: https://github.com/rootstrap/react-base/commit/b43e03eb66246b8143faa5ef6627dd81aa1d5ac1#diff-9968b87917a60f452c8ff987566a38e7 ? @agustito37

agustito37:
Yes it was, I needed the `process.env.GZIP_ENABLED` variable from `.env` so I decided to not use the webpack plugin and load them directly.

> Was this change intentional: [b43e03e#diff-9968b87917a60f452c8ff987566a38e7](https://github.com/rootstrap/react-base/commit/b43e03eb66246b8143faa5ef6627dd81aa1d5ac1#diff-9968b87917a60f452c8ff987566a38e7) ? @agustito37

webpack/webpack.config.prod.js
```diff
@@ -79,7 +78,9 @@ export default {
         test: /\.js$|\.css$/,
         threshold: 0,
         minRatio: 2
-      })
+      }),
+
+    new Dotenv({ path: envPath })
```
agustito37:
```suggestion
    new Dotenv({ path: envPath, systemvars: true })
```

peter7z: I would rather not expose system vars on a production build...

PR #191, opened 2020-04-06
Fix: Set headers when request data is null

nsantos16:
# Fix: Set headers when request data is null
### Description 
- Changed `applyDefaultInterceptors.js`  for set data as an empty object (`{}`), this allows set the content type even for empty requests.

### Docs 📚 
- [Why axios remove the headers for empty request? (line 119)](https://github.com/axios/axios/blob/dc4bc49673943e35280e5df831f5c3d0347a9393/lib/adapters/xhr.js)

### DIFF
--- a/src/httpClient/applyDefaultInterceptors.js
+++ b/src/httpClient/applyDefaultInterceptors.js
@@ -20,7 +20,7 @@ export default (store, client) => {
         uid
       };
     }
-    config.data = humps.decamelizeKeys(data);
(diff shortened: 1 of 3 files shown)


peter7z:
I think axios will not add 'data' for some methods.
https://github.com/axios/axios/blob/5e0fb5fc998d7f9f5a897363977b42f4fa047238/lib/core/Axios.js#L73

PR #192, opened 2020-04-13
Feature/add docker compose

mrubi-rootstrap:
I created a `Dockerfile` and a `docker-compose` file to create a local environment for a project from this base repository.
I thought it may be of use to somebody else, sharing it in this PR.


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Rootstrap React Redux Base",
   "engines": {
     "npm": ">4",
-    "node": "10.16.0"
+    "node": ">=10.16"
```
mato2593, line 7: I don't think that setting that the version should be greater or equal than the one we set is a good practice. This may create issues. We should either set a fixed version or use the caret (`^`) so that the major number is not updated and we ensure that there won't be breaking changes.

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+
+services:
+  frontend:
+    build:
+      .
```
juan-apa, line 6: this should be inline. Like: `build: .`

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:latest
```
juan-apa, line 1: Always use a version in the `FROM` statement. I see you updated the `package.json` to version 10.16. Here your statement should look like: `FROM node:10.16`

README.md
```diff
@@ -6,6 +6,20 @@
...
+### Build the image
+```
+docker-compose build
+```
+## Open a console on the container
+```
+docker-compose run \
+  -p 8000:8000 \
+  -v $(pwd)/src:/source_code/src \
+  -v $(pwd)/cypress:/source_code/cypress \
```
juan-apa, line 20: I'm pretty sure the usage of `pwd` variable here is unnecessary (The docker files live in the root of the project, so the build context is the root of the project. This means, that by mounting the volume like this: `src:/source_code/src` will suffice). What do you think about removing it to make the file more readable.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    libnotify-dev \
+    libgconf-2-4 \
+    libnss3 \
+    libxss1 \
+    libasound2 \
+    libxtst6 \
+    xauth \
+    xvfb
+COPY . /source_code
+WORKDIR /source_code
```
juan-apa, line 16: The convention when dockerizing a nodejs app is to put the application in `/usr/src/app`. You can read more about directoy hierarchy in linux [here](https://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/usr.html)

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
...
+    container_name: node-frontend
+    working_dir: /source_code
```
juan-apa, line 8: What do you think about adding a `node_modules` volume to cache your modules and not having to install them every-time your source code changes?
You can read more on how to achieve this [here](https://christianfei.com/posts/Cache-speed-up-docker-node-modules/)

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    libgtk2.0-0 \
+    libgtk-3-0 \
+    libnotify-dev \
+    libgconf-2-4 \
+    libnss3 \
+    libxss1 \
+    libasound2 \
+    libxtst6 \
+    xauth \
+    xvfb
```
juan-apa, line 14: Are all of these necessary? It seems like this are cypress requirements. However testing will not be done in production. What about splitting the images into a development and a production images. The production image should not include these packages, to optimize storage in the production server.

package.json
mrubi-rootstrap: I haven't thought it, thanks!
I used >= because since this is a template for new apps I did not want the version to get outdated. I am not sure how to keep this version up to date with the latest stable release of node, but I agree that leaving this in the template would create issues in the apps. I will change to using ^

docker-compose.yml
mrubi-rootstrap: Will fix it. Thanks!

Dockerfile
mrubi-rootstrap: Yes, thanks!
Since this repo is a template for new apps, Is there a way to flag that this number is not the most stable release to make us aware and update it?

README.md
mrubi-rootstrap: Will do. Thanks!

Dockerfile
mrubi-rootstrap: Did not know that, thanks! I will change it

docker-compose.yml
mrubi-rootstrap: Yes. Thanks!

Dockerfile
mrubi-rootstrap: They are indeed.
The first attempt was to make a development environment only. I know little about managing deployments but I will take a look how to split the Dockerfile for different targets.
Thanks!

mrubi-rootstrap:
@juan-apa @mato2593 Thanks for your reviews and links!

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+        libxss1 \
+        libasound2 \
+        libxtst6 \
+        xauth \
+        xvfb; \
+    fi
+
+WORKDIR /usr/src/app
+COPY . /usr/src/app
+RUN yarn install
```
juan-apa, line 22: What do you think about adding an `ENTRYPOINT` to the image and the `docker-compose.yml`?

juan-apa: Also, to effectively cache the dependencies, you should first copy the `package.json`, `package-lock.json` (`yarn.lock` if the project is using yarn), then install the dependencies using `npm install / yarn install` and finally copy the source code. That way the layers until the dependencies install get cached, and then the source code, which is what changes gets copied over. That way, the build time gets very very low.

mrubi-rootstrap: > What do you think about adding an `ENTRYPOINT` to the image and the `docker-compose.yml`?

Could be. The entry point would run `yarn build` or `yarn start` ?

mrubi-rootstrap: > Also, to effectively cache the dependencies, you should first copy the `package.json`, `package-lock.json` (`yarn.lock` if the project is using yarn), then install the dependencies using `npm install / yarn install` and finally copy the source code. That way the layers until the dependencies install get cached, and then the source code, which is what changes gets copied over. That way, the build time gets very very low.

Some npm installation scripts depend on other files than `package.json` and `package-json.log` so I just copied the whole dir but I understand your point. Thanks!

docker-compose-dev.yml
```diff
@@ -0,0 +1,16 @@
+version: '3'
+
+services:
+  frontend:
+    build:
+      context: .
+      args:
+        NODE_ENV: development
```
manuelcheveste, line 8: It is necessary to add this `NODE_ENV` arg and then on line 14 `- NODE_ENV = development`? Also, on the docker file I also see that you declare `NODE_ENV` arg. Shouldn't just keep the environment one?

docker-compose-dev.yml
```diff
@@ -0,0 +1,16 @@
...
+    container_name: node-frontend
+    working_dir: /usr/src/app
+    volumes:
+      - node_modules:/usr/src/app/node_modules
+    environment:
+      - NODE_ENV=development
+volumes:
+  node_modules: {}
```
manuelcheveste, line 16: why having this volume?

PR #199, opened 2020-06-04
Remove pwa support

### LINKED ISSUE #197: Remove PWA config

peter7z:
Tries to resolve #197
There might be something else that we can remove but I'm not really sure.
Still, this removes some stuff and everything, including SSR, still works,


webpack/webpack.config.prod.js
```diff
@@ -6,7 +6,6 @@ import WebpackMd5Hash from 'webpack-md5-hash';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
 import CompressionPlugin from 'compression-webpack-plugin';
-import { GenerateSW } from 'workbox-webpack-plugin';
```
fsarachu, line 9: `workbox-webpack-plugin` dependency should be removed from `package.json` too right?

PR #211, opened 2020-09-23
enhancement: upgrade useForm hook and add @rootstrap/validate

aguscha333:
### Description

This PR contains a couple of upgrades to the usage of forms. Here is a list:

* Add active state support for inputs, not being used in the base because the standard inputs don't need it but it's very useful for the material-design-like inputs that we usually use.
* Add the possibility to disable the submit button if the form is not valid, which required
	* Adding a useEffect hook in the input to register it in the values of useForm
	* Adding the touched flag so we can show errors only if the user has interacted with the field
	* In case the submit is not disabled when the form is invalid, touch all the fields on submit so the errors show
* Add a flag that lets you validate all the fields when one changes, useful if fields depend on others (i.e.: conditionalConstraints from @rootstrap/validate)
* Swapped validate.js for [Rootstrap's own validation lib](https://github.com/rootstrap/validate/) which is a wrapper of validate.js. It's a hub for cannon validate.js validators and also other custom validators that we have used that are so common that were calling to be grouped.

### Notes:
I updated the `.nvmrc` file to set the same version of node being set on the `package.json` otherwise the project wouldn't run, so that is fixed as well

### Preview

form

Since some of the functionalities added are not being used by the base but will be useful for our projects, here is a preview of how those functionalities would look like but taken from the react-native-base repo that is taking full advantage.

Here is a preview of how a form now would look like with active, touched, and disabled states.

useFormActiveAndTouched


lucia-santos (approved):
Nice!! 👏

felipegarcia92 (approved):
Thanks for working constantly on upgrading stuff 👏

PR #212, opened 2020-09-30
Add PR template

missx:
Adds PR template to the repo.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,8 @@
 <!--- This is an example of a PR template. You can adjust it to your project's needs. --->
 
-### [Type]: [Title of the card]
-
 <!---
+  PR Title suggestion - [Type]: [Title of the card]
+
   Types:
     feat: (new feature for the user, not a new feature for build script)
     fix: (bug fix for the user, not a fix to a build script)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,30 @@
+<!--- This is an example of a PR template. You can adjust it to your project's needs. --->
+
+### [Title of the card]
+
+## PR Type:
+- [ ] Feature
+- [ ] Bugfix
+- [ ] Refactoring
+- [ ] Documentation changes
+- [ ] Configuration changes
+- [ ] Other (please describe): 
+
+## Links (Jira/Trello ticket and other relevant links):
+<!--- At a minimum include links to the Jira/Trello ticket --->
+
+## What & Why:
+<!--- Describe the changes being made and why it's useful --->
+
+## Screenshots:
+<!--- Screenshots and videos of the new behavior are great to understand the PR --->
+
+## Testing steps:
+- [ ] Tested in different resolutions
+- [ ] Tested with SSR
+- [ ] Tested on Safari browser
+- [ ] Design reviewed with design team
+
+#### Notes:
+<!--- Any other relevant information --->
+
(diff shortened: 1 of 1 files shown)


nsantos16 (approved):

felipegarcia92 (approved):
👏 thanks for doing it

spereztoledo (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,30 @@
...
+
+### [Title of the card]
+
+## PR Type:
+- [ ] Feature
+- [ ] Bugfix
+- [ ] Refactoring
+- [ ] Documentation changes
+- [ ] Configuration changes
+- [ ] Other (please describe):
```
aguscha333, line 11: I'd suggest trying to adopt this methodology instead: [Semantic PR](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716). It will save space in the PR description and also it helps create a cleaner changelog.

It can be added as a check in the PRs to help ensure that it being followed:
image

We are currently using this in my project.

.github/pull_request_template.md
```diff
@@ -0,0 +1,30 @@
...
+- [ ] Other (please describe): 
+
+## Links (Jira/Trello ticket and other relevant links):
+<!--- At a minimum include links to the Jira/Trello ticket --->
+
+## What & Why:
+<!--- Describe the changes being made and why it's useful --->
+
+## Screenshots:
+<!--- Screenshots and videos of the new behavior are great to understand the PR --->
```
aguscha333, line 20: This section usually goes at the bottom and I think that part of the reason is that sometimes images take a lot of space, especially if they are vertical.

.github/pull_request_template.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Testing steps:
```
aguscha333, line 22: Hmm WDYT about calling it `Tasks`? `Testing steps` to me sounds like steps to reproduce or something like that.

aguscha333 (approved):
👏

.github/pull_request_template.md
missx: yes!

.github/pull_request_template.md
missx: Yeah, im MC we usually put them after the what & why, but I think at the bottom makes sense, since you can lose sight of what's below if not.

manema (approved):

.github/pull_request_template.md
missx: I changed the title to `[Type]: [Title]` as you suggested and left the different types as a comment for reference. Let me know what you think

.github/pull_request_template.md
```diff
@@ -1,25 +1,25 @@
...
-### [Title of the card]
-
-## PR Type:
-- [ ] Feature
-- [ ] Bugfix
-- [ ] Refactoring
-- [ ] Documentation changes
-- [ ] Configuration changes
-- [ ] Other (please describe): 
+### [Type]: [Title of the card]
```
aguscha333, line 3: We usually don't put this line and just use the PR title as the title. I would move this as a comment as well. But I leave this up to you since it can be more of a personal preference than anything else.

aguscha333 (approved):

.github/pull_request_template.md
missx: Got it! I would prefer to have what's used the most, so I'll move it to a comment.

PR #217, opened 2020-10-30
fix: reload page on SSR

elizabethlofredo:
## What & Why:
   - Fix  SSR reload private pages: when requesting a private page with SSR, the server has no information about authentication so the result was a redirect to the home page.
  - Change `yarn ssr`  to use `.env.dev` as default environment. The same way`yarn start` does, so it's consistent. (If another environment is needed we use ENV='enviroment' as usual)
  - Also update _react-helmet_ version that was triggered a warning. 

   **Context**: 
Previously when we had `redux-react-session` on the base, we had an "initSession" function that loads the session from the cookies, when we remove this library to use `redux-persist` we lost that part needed for SSR.
In this PR I added the logic to store the session on the cookies, in this way the server loads the cookies from the request and has information about the session. 

Any comments/suggestions or changes on how this was implemented are welcome 😄 

## Tasks:
- [x] Tested with SSR

## Screenshots:
**Before:**

gif

------

**After:**  
gif


server/server.js
```diff
@@ -32,8 +36,23 @@ server
...
+    let sessionInfo;
+    try {
+      const cookieJar = new NodeCookiesWrapper(new Cookies(req, res));
+      const config = {
+        ...sessionPersistConfig,
+        storage: new CookieStorage(cookieJar),
+      };
+      sessionInfo = await getStoredState(config);
+    } catch (error) {
+      console.log(error)
```
camiriveron, line 50: Should we do something else with this error? Would the store just initialize with an empty state in this case?

camiriveron (approved):
Nice one! 👌 Left one small comment

agustito37 (approved):
I do not have enough knowledge of this but seems good to me.

aguscha333 (approved):

nsantos16 (approved):

server/server.js
elizabethlofredo: Yes, the store will initialize with an empty state. Not sure what we should do with the error... suggestions? 😬

server/store/index.js
```diff
@@ -0,0 +1,20 @@
+import { createStore, compose, applyMiddleware } from 'redux';
+import { thunkMiddleware } from '@rootstrap/redux-tools';
```
missx, line 2: Is there urgency to merge this PR? I see you're using `redux-tools` here. If we merge [this one](https://github.com/rootstrap/react-base/pull/213) first you'll have to update here. Otherwise we'll have to update the other PR.

elizabethlofredo: You are right, maybe will be better to merge the other PR first and I can update this one after

server/server.js
missx: I'm not sure either 😂 Maybe as a recommendation I would use `console.error` instead of log to better display the error itself.

PR #219, opened 2020-12-02
allow logout method to work when server responds with unauthorized error

alexfradiani:
## What & Why:
Currently, when the logout action gets called, if the server responds with an error status, an uncontrolled loop can be triggered, and the authenticated reset state never happens.

With this change the client status will be forced to be logged out, which is a more desirable state in the react side regardless of what happened in the server.

### DIFF
--- a/src/httpClient/applyDefaultInterceptors.js
+++ b/src/httpClient/applyDefaultInterceptors.js
@@ -41,7 +41,7 @@ export default (store, client) => {
     },
     error => {
       if (error.response && error.response.status === UNAUTHORIZED) {
-        store.dispatch(logout());
+        store.dispatch(logout.success());
       }
 
       return Promise.reject(error);


src/httpClient/applyDefaultInterceptors.js
```diff
@@ -8,7 +8,7 @@ const CLIENT = 'client';
 const UNAUTHORIZED = 401;
 
 export default (store, client) => {
-  client.interceptors.request.use(config => {
+  client.interceptors.request.use((config) => {
```
martinjaimem, line 11: Just curious, do we use parenthesis around a single-argument function?

src/httpClient/applyDefaultInterceptors.js
```diff
@@ -17,31 +17,31 @@ export default (store, client) => {
         ...headers,
         [ACCESS_TOKEN]: token,
         client,
-        uid
+        uid,
```
martinjaimem, line 20: Same here with the comma

src/httpClient/applyDefaultInterceptors.js
aguscha333: nope

src/httpClient/applyDefaultInterceptors.js
aguscha333: This depends on the project but I'd assume that this was as it was because of the eslint config so I'd leave it as it was.

alexfradiani: yes, this was a prettier config that got in the way from my local environment. Will reupload in a while with the previous formatting.

agustito37 (approved):

martinjaimem:
I'm closing this one due to lack of activity 😢

PR #221, opened 2020-12-18
adding reset action in redux toolkit set up

manema:
## What & Why:
We need a way to clean an action status from the status reducer. Generally, we use actions status to show errors or move to other pages. 
For instance, if we have a component which performs a request and it's rejected, in most cases we would like to show a message then. If we go to other page and then come back to the one with this component we will see the error again because it's stored in the status reducer.

## Tasks:
- [ ] Tested in different resolutions
- [x] Tested with SSR

## Notes:
**Example**
we can use it as follows:
`const loginReset = useDispatch(login.reset);`

### DIFF
--- a/src/state/reducers/statusReducer.js
+++ b/src/state/reducers/statusReducer.js
@@ -1,6 +1,6 @@
 import { createReducer } from '@reduxjs/toolkit';
 
-import { FULFILLED, REJECTED, PENDING } from 'constants/actionStatusConstants';
+import { FULFILLED, REJECTED, PENDING, RESET } from 'constants/actionStatusConstants';
 
 const DELIMITER = '/';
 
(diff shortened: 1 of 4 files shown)


missx (approved):

mariosouto (approved):

nsantos16 (approved):

elizabethlofredo (approved):
Great, thanks Ema 🙌

src/state/reducers/statusReducer.js
```diff
@@ -30,5 +30,12 @@ export default createReducer({}, builder => {
         state[getActionKey(type)] = { status: PENDING };
       }
     )
+    .addMatcher(
+      ({ type }) => type.endsWith(`/${RESET}`),
+      (state, { type }) => {
+        delete state[getActionKey(type)];
```
agustito37, line 36: Fun fact: `delete obj.prop` is less performant that `obj.prop = undefined`, at least in v8.

agustito37 (approved):
Good job!

PR #224, opened 2021-02-11
fix: GZIP_ENABLED flag

andrescorrea:
# Fix GZIP_ENABLED flag 

## What & Why:

As the `GZIP_ENABLED` flag is mandatory (that is, if it is not set in the `.env`, the build does not work), a default value is added to s3 tools in order to prevent the build fail if it's not set.

Also `GZIP_ENABLED` is added to the `.env.example` table in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 2. Install dependencies: `yarn` or `npm install`
 3. Create the environment variables files in root folder(.env.dev, .env.staging and .env.prod):
 
-`.env.example` example:
+`.env.example` [example](/.env.sample):
 
 ```
   API_URL=http://your-api-url.com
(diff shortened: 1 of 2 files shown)


nsantos16 (approved):
Awesome 🚀

README.md
```diff
@@ -20,14 +20,15 @@
...
-`.env.example` example:
+`.env.example` [example](/.env.sample):
 
 ```
   API_URL=http://your-api-url.com
   AWS_BUCKET=bucket
   AWS_REGION=region
   AWS_ACCESS_KEY_ID=key_id
   AWS_SECRET_ACCESS_KEY=secret_key
+  GZIP_ENABLED=false
```
nsantos16, line 31: Just as a suggestion, maybe we can create a table with the different values for the environment using a table, check this [page](https://www.tablesgenerator.com/markdown_tables) for generating one of them.  I think the columns will be "Environment key", "Required", "Default", WDYT?

PR #226, opened 2021-03-15
Remove enzyme and unused testing utils

nsantos16:
## What & Why:

Removed `enzyme` dependency and other unused files related to tests.

## Tasks:
- [x] Tested with SSR
- [ ] Tested on Safari browser

### DIFF
--- a/tools/testSetup.js
+++ b/tools/testSetup.js
@@ -1,6 +1,0 @@
-import Enzyme from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
-
-require('dotenv').config({ path: '.env.test'});
-
-Enzyme.configure({ adapter: new Adapter() });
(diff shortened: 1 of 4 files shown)


kevquincke (approved):
Finally! 👏

tools/testSetup.js
```diff
@@ -1,6 +0,0 @@
-import Enzyme from 'enzyme';
```
JuAnmdq, line 1: We can take advantage of `tools/testSetup.js` that will render at setupFilesAfterEnv to import only once at start `import '@testing-library/jest-dom';` in order to not import in every file common functionalities like `.toBeInTheDocument()`

nsantos16: Agree! I will introduce these changes in another PR related to react testing library 😄

nsantos16: WDYT about using the next folder structure for that `__tests__/setup/preEnv.js` ?

JuAnmdq: In this case, we should move all jest / tl testing related's tools in that folder, is correct?

PR #227, opened 2021-03-18
Enhancement: React testing library and accessibility linters

nsantos16:
## What & Why:
The aim of this PR is to restructure the way that we make testing in RS, introducing [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) as a tool and restructuring Cypress folder structure on the road. Also, some tools like [test-data-bot](https://github.com/jackfranklin/test-data-bot) and [msw](https://mswjs.io/) are integrated to facilitate testing processes.

Additionally, following the best practices of RTL([queries with roles and label text are at the top of the priority](https://testing-library.com/docs/queries/about/)) and accessibility practices, I integrated `eslint-plugin-jsx-a11y` to prevent components that don't follow the a11y standard.

## Tasks:
- [x] Tested with SSR
- [ ] Tested on Safari browser

## Resources 📚:
- [Stop mocking fetch](https://kentcdodds.com/blog/stop-mocking-fetch/) -> msw
- [Avoid nesting when you're testing](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) -> don't use it and before all every time
- [Write tests, not too many, mostly integration](https://kentcdodds.com/blog/write-tests/)


JuAnmdq:
Integration tests are really fast for executing, we can add in `lint-staged` after `eslint-fix` and before `git add` to prevent push when some test fails on pre-commit

.eslintrc
```diff
@@ -5,7 +5,10 @@
     "plugin:import/warnings",
     "airbnb",
     "prettier",
-    "prettier/react"
+    "prettier/react",
+    "plugin:jest-dom/recommended",
+    "plugin:testing-library/react",
+    "plugin:jsx-a11y/recommended"
```
JuAnmdq, line 11: You shouldn't put also `jest-dom` and `testing-library` as **plugins**?

JuAnmdq, line 11: Can we add also "plugin:testing-library/recommended" as extension?

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  test('Testing Login submit', async () => {
+    const emailField = screen.getByLabelText(/email/i);
+    const passwordField = screen.getByLabelText(/password/i);
+    const onSubmitLogin = screen.getByRole('button', { name: /submit/i });
+
+    fireEvent.click(onSubmitLogin);
+    const alerts = screen.getAllByRole('alert');
+    expect(alerts).toHaveLength(2);
```
JuAnmdq, line 36: Instead of this, we can expect the constraints required messages to be more specific in this case:
expect(screen.getByText(/You must enter an email to continue/)).toBeInTheDocument();
expect(screen.getByText(/You must enter a password to continue/)).toBeInTheDocument();

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+    fireEvent.change(emailField, { target: { value: user.email } });
+    fireEvent.change(passwordField, { target: { value: user.password } });
+    fireEvent.click(onSubmitLogin);
+    screen.getByRole('status');
+
+    await waitFor(() =>
```
JuAnmdq, line 43: We can consider this Kent. C. Dodds recommendation about `waitFor`:
[https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find)

tests/rtl/integration/LoginPageRTL.spec.js
nsantos16: Agree, thx!

tests/rtl/integration/LoginPageRTL.spec.js
nsantos16: yep, I forgot to remove this `waitFor` (I had some problems with circleCI and for debugging I used it), thx for notice it 😄

tests/cypress/stubs/sessionStubs.js
```diff
@@ -16,7 +16,7 @@ export const loginStub = customUser => ({
   cases: {
     success: {
       status: SUCCESS_CODE,
-      response: { user: customUser || user() }
+      response: { user: customUser || userBuilder() }
```
natokra, line 19: This way I think this might be a better way to evaluate if customUser is undefined or null, otherwise it might have some trouble resolving that scenario.

```suggestion
      response: { user: customUser != null ? customUser : userBuilder() }
```

tests/cypress/stubs/sessionStubs.js
```diff
@@ -33,7 +33,7 @@ export const signUpStub = customUser => ({
   cases: {
     success: {
       status: SUCCESS_CODE,
-      response: { user: customUser || user({ complete: false }) }
+      response: { user: customUser || userBuilder({ complete: false }) }
```
natokra, line 36: same case than before

natokra (commented):
this little thing is the only one that I found. If you consider that is not something to change go ahead!

.eslintrc
nsantos16: Screen Shot 2021-03-19 at 1 05 43 PM
According to the documentation `plugin:testing-library/react` extends of `plugin:testing-library/recommended`

nsantos16: I added `import '@testing-library/jest-dom';` on `setUpTests.js` for import jest-dom with testing-library specifications

natokra (approved):

.eslintrc
JuAnmdq: I think `@testing-library/jest-dom` is about functionality, but you can try with some eslint rules exception from both `jest-dom` and `testing-library`, if its alerts some package rules of docs, should be work fine.

JuAnmdq: fair enough! good to know

ManuViola77 (approved):
Awesome 👏🏻

tests/rtl/mockServices/userMockServices.js
```diff
@@ -0,0 +1,21 @@
+import { rest } from 'msw';
+import faker from 'faker';
+
+import userFactory from 'fixtures/userFactory';
+
+const mockLoginRequest = rest.post(/\/users\/sign_in/i, async (req, res, ctx) => {
```
JuAnmdq, line 6:
```suggestion
const mockLoginRequest = rest.post('/users/sign_in', async (req, res, ctx) => {
```

Be careful here, this regex compare if contains this paths, but if you call in real services to `/foo/users/sign_in/bar` it's going to intercept that route this test its going to work anyway, in the future this will pass many tests that call another routes. Is more secure to compare just with a little bit strict string.

For example if you have another different post service `/accounts/users/sign_in` `msw` it going to intercept this one.

Also it will be a good idea when no route is matching with for example at end of this handlers, match res.post(*, (req, res, ctx) => { response and throw and exception here }) in order to know whats going on when route is not matching

webpack/shared/resolve.js
```diff
@@ -16,9 +16,8 @@ module.exports = {
     selectors: path.resolve(__dirname, '../../src/selectors'),
     styles: path.resolve(__dirname, '../../src/styles'),
     utils: path.resolve(__dirname, '../../src/utils'),
-    fixtures: path.resolve(__dirname, '../../cypress/fixtures'),
-    stubs: path.resolve(__dirname, '../../cypress/stubs'),
-    cypressConstants: path.resolve(__dirname, '../../cypress/support/constants'),
-    reusableTests: path.resolve(__dirname, '../../cypress/support/reusableTests')
+    fixtures: path.resolve(__dirname, '../../tests/fixtures'),
+    cypress: path.resolve(__dirname, '../../tests/cypress'),
+    rtl: path.resolve(__dirname, '../../tests/rtl')
```
JuAnmdq, line 21: this resolve is also will configure in `webpack.config.prod`, is really necessary this config there?

tests/rtl/mockServices/mockServer.js
```diff
@@ -0,0 +1,10 @@
+import { rest } from 'msw';
+import { setupServer } from 'msw/node';
+
+import userMockServices from 'rtl/mockServices/userMockServices';
+
+const serverHandlers = [...userMockServices];
```
JuAnmdq, line 6:
```suggestion
const handlers = [...userMockServices];
```
**Trivial:** you can just call it handlers since you are already in mockServer file, and it is a litte bit redundant

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  test('Testing Login submit', async () => {
+    const emailField = screen.getByLabelText(/email/i);
+    const passwordField = screen.getByLabelText(/password/i);
+    const onSubmitLogin = screen.getByRole('button', { name: /submit/i });
+
+    fireEvent.click(onSubmitLogin);
+    expect(screen.getByText(/You must enter an email to continue/)).toBeInTheDocument();
+    expect(screen.getByText(/You must enter a password to continue/)).toBeInTheDocument();
+
+    fireEvent.change(emailField, { target: { value: user.email } });
```
JuAnmdq, line 38:
```suggestion
    userEvent.type(emailField, user.email);
```

Can we start using the recommended for Kent C. Dodds `userEvent` that also consider `keyDown`, `keyPress`, and `keyUp` events and also it looks like a little bit less boilerplate

[https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event)

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { axe } from 'jest-axe';
+import { screen, fireEvent, waitFor } from '@testing-library/react';
```
JuAnmdq, line 4:
```suggestion
import { screen, fireEvent } from '@testing-library/react';
```
waitFor is unused in this file, but beyond that, I notice that `eslint --fix` that is implement on `lint-staged` is not already consider `webpack` and `tests`, maybe you can add `tests` to existing `lint:code` and then call `yarn lint` in `lint-staged` with `run-script` or the workaround that lint-staged need to run local script.
In this way before each commic you will lint `src`, `webpack` and `tests` folders and  `style-lint` instead of only `src` folder

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
```
JuAnmdq, line 1: This file can be called just `LoginPage.spec.js` since it already on `rtl/integration` folder, it will be a little bit longer route with long filename.

nsantos16: There's another file on cypress that adopts that name 😞

tests/rtl/integration/LoginPageRTL.spec.js
nsantos16: Nice catch!

tests/rtl/mockServices/userMockServices.js
nsantos16: I have some problems with Circle CI because the origin of the request is random (i.e 'circle-repo/users/sign_in'), I have to figure out how to solve that problem in an efficient way

webpack/shared/resolve.js
nsantos16: Yes it necessary for include when the application runs (cypress), also if we are gonna do E2E testing with cypress will we need this config

tests/rtl/mockServices/userMockServices.js
JuAnmdq: maybe enforce some env variable for the origin? it just an idea

tests/rtl/integration/LoginPageRTL.spec.js
JuAnmdq: But there are in different folders, anyways both ways are ok

tests/rtl/mockServices/userMockServices.js
nsantos16: Fixed, it was the circle CI variable for the API URL. For change that I deleted from circle CI the variable and change a little bit `config.yml` to set `.env.test` in cypress running

JuAnmdq (approved):

tests/rtl/integration/LoginPageRTL.spec.js
```diff
@@ -0,0 +1,53 @@
...
+import userFactory from 'fixtures/userFactory';
+import { render } from 'rtl/utils';
+
+describe('Testing Login page submit', () => {
+  const user = userFactory();
+  let wrapper;
+
+  beforeEach(() => {
+    wrapper = render(
+      <MemoryRouter initialEntries={['/login']}>
```
JuAnmdq, line 18: `initialEntries` is currently working? Did you try with another route like `/sign-up`?

PR #230, opened 2021-04-08
fix: set browser locale within Cypress tests

nachosoloaga:
## What & Why:

Added language property set to english to browser within Cypress, so tests that expect certain strings in html tags don't fail because of internationalization issues. E.g. a test that expects the string 'Login' (as hardcoded in the code) will encounter the internationalized string 'Iniciar sesión' because of the browser default locale, causing the test to fail. Now the browser default locale will always be english.

## Screenshots:

- Testing output before fix
image

- Testing output after fix
image

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -24,6 +24,7 @@ addMatchImageSnapshotCommand({
 });
 
 Cypress.on('window:before:load', win => {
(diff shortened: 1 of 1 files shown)


felipegarcia92 (approved):
👏  thanks!

Nice PR description too 👌

aguscha333 (approved):
🚀

PR #232, opened 2021-04-16
changes on the App component in order to use one file

rodrigorec:
## What & Why:

Changes on the App component to use only one file.

## Tasks:
- [ ] Tested in different resolutions
- [x] Tested with SSR


matteo95g (approved):

aguscha333 (approved):

src/components/App.js
```diff
@@ -4,27 +4,42 @@ import { Helmet } from 'react-helmet';
...
 
+  const RenderRoutes = ({ children }) => {
+    return !isSSR ? <BrowserRouter>{children}</BrowserRouter> : <>{children}</>;
+  };
+
   return (
     <>
       <Helmet>
-        <title>RS React Redux Base</title>
+        <title>{title}</title>
```
JuAnmdq, line 21: Is `title` really need to pass as a prop in App?
Plus: a better approach using `helmet` is set specific title for each page, it will help to avoid duplicated title and match SEO with search intent content. WDYT?

